PR #7, opened 2017-11-15
Fix patch mysql context

clark-pan:
Not a massive issue, but the context for the patching functions are wrong.


haotianw465:
Hi, could you explain a little bit more about this change?

clark-pan:
Non-issue, the underlying code doesn't care what `thisArg` is at all.

PR #11, opened 2017-12-04
WIP: work around cls handling of Promises

omsmith:
continuation-local-storage chooses to bind the context to the Promise
constructor and passes this to any and all reactions to that promise.
While this might make sense in some cases, it very much doesn't when
it comes to tracing an HTTP request context. Provide a patcher for
Promises which will bind the reaction handler to the active segment
when "then" or "catch" is called.

Hey X-Ray team

Didn't expose this patcher or go through the work of integrating it, but wanted to write up a test exposing the issue, and share a possible solution to it.

In the current state, if a Promise is created in request/segment A, and waited upon in request/segment B, than all metadata/errors/subsegments etc. after the fact will be added to segment A instead of B. This is due to the way continuation-local-storage opts to bind Promises, leaning more strictly towards execution paths, which I don't believe fits for a product like X-Ray.

Certainly not the first APM-style product to have this problem, since they all use cls :)

Let me know your thoughts, I'll point our TAMs this way as well.

- Owen

### DIFF
--- a/packages/core/lib/patchers/promise_p.js
+++ b/packages/core/lib/patchers/promise_p.js
@@ -1,0 +1,35 @@
+/**
+ * @module promise_p
+ */
+
+/**
+ * This module patches Promise libraries to correct tracing attribution for http requests.
+ */
+
+var contextUtils = require('../context_utils');
+
+function patchPromise(Promise) {
+  var then = Promise.prototype.then;
+  Promise.prototype.then = function(onFulfilled, onRejected) {
+    if (contextUtils.isAutomaticMode()
+      && tryGetCurrentSegment()
+    ) {
+      var ns = contextUtils.getNamespace();
+
+      onFulfilled = onFulfilled && ns.bind(onFulfilled);
+      onRejected = onRejected && ns.bind(onRejected);
+    }
+
+    return then.call(this, onFulfilled, onRejected);
+  };
+}
+
+function tryGetCurrentSegment() {
+  var segment = null;
+  try {
+    segment = contextUtils.getSegment();
+  } catch(e) { /**/ }
+  return segment;
+}
+
+module.exports = patchPromise;
--- a/packages/core/test/integration/segment_maintained_across_shared_promise.test.js
+++ b/packages/core/test/integration/segment_maintained_across_shared_promise.test.js
@@ -1,0 +1,73 @@
+/* global Promise */
+
+if (!global.Promise) {
+  process.exit(0);
+}
+
+var assert = require('chai').assert;
+var http = require('http');
+
+var AWSXRay = require('../../');
+var Segment = AWSXRay.Segment;
+
+require('../../lib/patchers/promise_p')(Promise);
+
+AWSXRay.enableAutomaticMode();
+
+var sharedPromise = null;
+
+var server = http
+  .createServer(function(req, res) {
+    var ns = AWSXRay.getNamespace();
+    ns.bindEmitter(req);
+    ns.bindEmitter(res);
+
+    ns.run(function () {
+      var segment = new Segment('foo');
+
+      AWSXRay.setSegment(segment);
+
+      if (!sharedPromise) {
+        sharedPromise = Promise.resolve();
+      }
+
+      sharedPromise.then(function() {
+        var retrievedSegment = AWSXRay.getSegment();
+        res.end();
+
+        // setTimeout so the assertion isn't caught by the promise
+        setTimeout(function() {
+          assert.equal(segment.id, retrievedSegment.id);
+        });
+      });
+    });
+  }).listen(0, function() {
+    var address = server.address();
+
+    var count = 0;
+    function cb(err) {
+      if (err) {
+        throw err;
+      }
+
+      if (++count === 2) {
+        server.close();
+      }
+    }
+    sendRequest(address, cb);
+    sendRequest(address, cb);
+  });
+
+function sendRequest(address, cb) {
+  http
+    .request({
+      hostname: address.address,
+      port: address.port,
+      path: '/'
+    })
+    .on('response', function(res) {
+      res.on('end', cb).resume();
+    })
+    .on('error', cb)
+    .end();
+}


packages/core/lib/patchers/promise_p.js
```diff
@@ -0,0 +1,35 @@
...
+      onRejected = onRejected && ns.bind(onRejected);
+    }
+
+    return then.call(this, onFulfilled, onRejected);
+  };
+}
+
+function tryGetCurrentSegment() {
+  var segment = null;
+  try {
```
omsmith: `try/catch` because there's no internal function that just does `ns.get(SEGMENT)`

packages/core/lib/patchers/promise_p.js
```diff
@@ -0,0 +1,35 @@
...
+ * This module patches Promise libraries to correct tracing attribution for http requests.
+ */
+
+var contextUtils = require('../context_utils');
+
+function patchPromise(Promise) {
+  var then = Promise.prototype.then;
+  Promise.prototype.then = function(onFulfilled, onRejected) {
+    if (contextUtils.isAutomaticMode()
+      && tryGetCurrentSegment()
```
omsmith: Only do if inside of an active segment

packages/core/test/integration/segment_maintained_across_shared_promise.test.js
```diff
@@ -0,0 +1,73 @@
...
+  process.exit(0);
+}
+
+var assert = require('chai').assert;
+var http = require('http');
+
+var AWSXRay = require('../../');
+var Segment = AWSXRay.Segment;
+
+require('../../lib/patchers/promise_p')(Promise);
```
omsmith: Test doesn't fail because the patcher is in place

haotianw465:
Sorry for the late response. I took a look at your code. I understand your use case it makes sense to me. The patcher code is not tied to any particular use case and it will affect all `then` call if patched. 

From my understanding this could cause some side effect. If a segment A is waited on subsegment B, do subsequent subsegments/annotations/metadata get added to subsegment B?

We are trying to understand the general use case and downside of the change so we can provide proper guidance on when this patch should be used.

omsmith:
Just ate dinner and have to head to the store, but I'll have a response before I go to bed.

omsmith:
Hey @haotianw465 

I'm writing up some examples, but can you elaborate on what you mean by "If a segment A is waited on subsegment B"? Not sure I 100% understand what you're meaning, and would like to be able to address it fully.

Thanks!

omsmith:
Alright, here's what I have so far.

Would like to add examples using subsegments, but I would want to actually verify it against x-ray rather than just typing whatever I want into an editor and pretending I'm right. Similarly, the examples I've provided, I have not exactly verified at this point in time, but I feel pretty confident about them.

I would have double-checked and verified before replying, but it's almost 1:30a and I should head to bed.

If you can find time to do some of this investigation as well or instead, that would obviously be much appreciated (and I assume you wouldn't want to take my word for it anyway).

# How Promises currently work

To start off with, this is approximately how Promises currently work with continuation-local-storage (or async-listener). Simplifying a bit due to the complexities back and forth between the two, but I think it should at least show how it impacts x-ray:

```js
var Promise = global.Promise;
function WrappedPromise(executor) {
	var contextAtInitialConstruction = currentContext;

	var thisArg, resolve, reject;
	// Get resolve and reject from a real promise
	var promise = new Promise(getResolveRejectArgs);
	promise.prototype = WrappedPromise;

	// save the context that was active when the Promise was created so we can restore is later
	promise._context = contextAtInitialConstruction;

	// This is what normally happens during new Promise(), approximately
	try {
		executor.call(thisArg, resolve, reject);
	} catch (e) {
		reject(e);
	}

	return promise;

	function getResolveRejectArgs(_resolve, _reject) {
		thisArg = this;
		resolve = _resolve;
		reject = _reject;
	}
}

var then = Promise.prototype.then;
WrappedPromise.prototype.then = Promise.prototype.then = function(onResolve, onReject) {
	var self = this;

	// the returned promise will similarly propagate our initial context through it's then calls
	var promise = then.call(this, bindToContext(onResolve), bindToContext(onReject));
	promise._context = this._context;
	return promise;

	function bindToContext(fn) {
		if (typeof fn !== 'function') return fn;

		return function() {
			// set the active context to the one that was active when the promise was created
			currentContext = self._context;

			return fn.call(this, arguments);
		}
	}
};

global.Promise = WrappedPromise;
```

So, whenever our functions to `.then()` are called, they'll be put back to the context that was active when `new Promise()` was called, as opposed to the context that was active when we called `.then()`.

The reason it behaves this way is because it leans more closely to providing a stack trace. Promises kind of create two execution paths though - the path of actual execution (the stack trace), and the path created by calling then. This is a similar case to EventEmitters, and why we have to call `ns.bindEmitter()` on them. it's fortunate that bindEmitter works the way we would expect for an http tracing library though. The context is saved when `on` is called, and restored when the listener is called. For Promises, `then` maps pretty well to `on`, so the current behaviour would be like if the EE patching saved the state when the EE is created, or perhaps when emit was called.

# Examples

Each of these examples is using express, although I've omitted the boilerplate / setup you would do to get x-ray going with it.

## Example 1: No shared promises

This example will behave the same before and after the change. All promises are created in the same context then is called in, so they're equivalent. If an application's code looked like this, they would see no change.

```js
const http = require('http');

function workA() {
	return new Promise((resolve, reject) => {
		http
			.get('http://example.com/a', resolve)
			.on('error', reject);
	});
}

function workB() {
	return new Promise((resolve, reject) => {
		http
			.get('http://example.com/b', resolve)
			.on('error', reject);
	});
}

function workFoo() {
	return new Promise((resolve, reject) => {
		http
			.get('http://example.com/foo', resolve)
			.on('error', reject);
	});
}

function workBar() {
	return new Promise((resolve, reject) => {
		http
			.get('http://example.com/bar', resolve)
			.on('error', reject);
	});
}

app.get('/foo', (req, res) => {
	workA()
		.then(workB)
		.then(workFoo)
		.then(() => res.end());

});


app.get('/bar', (req, res) => {
	workA()
		.then(workB)
		.then(workBar)
		.then(() => res.end());

});
```

We'll call the endpoints ~concurrently

```sh
curl http://localhost/foo &
curl http://localhost/bar &
```

In x-ray, we should see two traces (not actually verified, because it's late)

* Trace 1
	* HTTP a
	* HTTP b
	* HTTP foo
* Trace 2
	* HTTP a
	* HTTP b
	* HTTP bar

## Example 2: Caching results as a Promise

One use-case you might have for sharing a promise is for caching, and simply storing the promise to the result. This is where we'll start seeing differences.

We'll update workB to cache the returned result.

```js
// I explicitly wrote this _not_ to also dedupe, so it looks a little weird.
function workB() {
	let cachedPromise = null;

	return function() {
		if (cachedPromise) {
			return cachedPromise;
		}

		const promise = new Promise((resolve, reject) => {
			http
				.get('http://example.com/b', resolve)
				.on('error', reject);
		}).then(result => {
			cachedPromise = promise;
			// clear the cache value later
			setTimeout(() => cachedPromise = null, 60 * 60 * 1000).unref()

			return result;
		});
	};
}
```

This time we'll call it serially to be sure the cached value is in place, but really this just shows how it might "only happen sometimes" in an application using this approach.

```sh
curl http://localhost/foo
curl http://localhost/bar
```

In x-ray, we should once again see two traces

* Trace 1
  * HTTP a
  * HTTP b
  * HTTP foo
  * HTTP bar
* Trace 2
  * HTTP a

Trace 1 received the work for "bar", even though `/foo` takes no part in making that work happen, other than having constructed the initial cached Promise value for "b". There may be some nuances here in terms of trace 1 having already been ended, but for sure it will be missing from trace 2.

With the patcher in place, we should see:

* Trace 1
  * HTTP a
  * HTTP b
  * HTTP foo
* Trace 2
  * HTTP a
  * HTTP bar

### Refactor to avoid it

In this case, it's pretty easy to change the code to not actually share the Promise.

```js
function workB() {
	let cachedValue = null;

	return function() {
		if (cachedValue) {
			return Promise.resolve(cachedValue);
		}

		new Promise((resolve, reject) => {
			http
				.get('http://example.com/b', resolve)
				.on('error', reject);
		}).then(result => {
			cachedValue = result;
			// clear the cache value later
			setTimeout(() => cachedValue = null, 60 * 60 * 1000).unref()

			return result;
		});
	};
}
```

## Example 3: Deduping requests

Another use-case that comes up is sharing a promise while some amount of work is in progress, in order to easily dedupe it.

```js
function workB() {
	let inflightRequest = null;

	return function() {
		if (!inflightRequest) {
			inflightRequest = new Promise((resolve, reject) => {
				http
					.get('http://example.com/b', resolve)
					.on('error', reject);
			}).then(result => {
				inflightRequest = null;
				return result;
			}, error => {
				inflightRequest = null;
				throw error;
			});
		}

		return inflightRequest;
	};
}
```

Calling it ~concurrently:

```sh
curl http://localhost/foo &
curl http://localhost/bar &
```

And once again, in x-ray, we'll see two unfortunate traces:

* Trace 1
  * HTTP a
  * HTTP b
  * HTTP foo
  * HTTP bar
* Trace 2
  * HTTP a

With the patcher in place, we should see:

* Trace 1
  * HTTP a
  * HTTP b
  * HTTP foo
* Trace 2
  * HTTP a
  * HTTP bar

### Refactoring to avoid it

This one is a bit trickier to refactor, and involves writing a bunch of code you normally wouldn't just to please x-ray

```js
function workB() {
	let inflight, waitingResolves, waitingRejects;

	reset();

	function reset() {
		inflight = false;
		waitingResolves = [];
		waitingRejects = [];
	}

	return function() {
		const promise = new Promise((resolve, reject) => {
			waitingResolves = resolve;
			waitingRejects = reject;
		});

		if (inflight) {
			return promise;
		}

		inflight = true;

		http
			.get('http://example.com/b', res => {
				waitingResolves.forEach(r => r(res));
				reset();
			})
			.on('error', e => {
				waitingRejects.forEach(r => r(e));
				reset();
			});

		return promise;
	};
}
```

# Manual mode

Some of this is fixable in manual mode, but some of it may not be. Also not a fan of polluting our code bases with the manual wireup.

haotianw465:
Thank you for the writeup. It is very useful information. We will also do some testing to verify the behavior with/without your patch. 

Sharing work across requests is definitely a common use case and we are trying to provide as simple instrumentation as we can for X-Ray customers. We will prioritize this fix so that no customer has to write extra code changes to make it work.

One thing I do want to clarify is that would you prefer to work with us to refine your PR at a state it can be directly merged-in, or you would like us to take the inspiration and just release a fix?

omsmith:
My (personal, and company) priority would be seeing it merged in and released, so if your team has the cycles to make that happen without me, would prefer that happen to avoid any delays in that back and forth (such as me taking 3 day to respond to this :smile:).

If that's the case, I would simply appreciate if you could use my commit as a basis for your work, if it's at all usable.

haotianw465:
I did some initial testing for `Example 1: No shared promises`. I'm trying to clarify a little bit of what you said and what I found.

Here is the hierarchy without your promise patch code:

* Trace1
	* WorkA
		* WorkB
  			* WorkFoo

Here is the hierarchy with your patch code:

* Trace1
	* WorkA
	* WorkB
	* WorkFoo

The SDK was development before native Promise support, and we had not tested Promise chains before. The above results show that the way CLS resolves context for a chain of promises doesn't fit the X-Ray use case well. 

I believe the patch code you provided fit for all use cases under tracing context. I will continue to do more testing on other cases. But I would like to confirm what I observed above is aligned with what you expect or what you saw from your testing code.

omsmith:
Safest thing might be to translate it as callbacks and compare?

Would be... approximately this?

```
const http = require('http');

function workA(cb) {
	http
		.get('http://example.com/a', r => cb(null, r))
		.on('error', cb);
}

function workB(cb) {
	http
		.get('http://example.com/b', r => cb(null, r))
		.on('error', cb);
}

function workFoo(cb) {
	http
		.get('http://example.com/foo', r => cb(null, r))
		.on('error', cb);
}

function workBar() {
	http
		.get('http://example.com/bar', r => cb(null, r))
		.on('error', cb);
}

app.get('/foo', (req, res) => {
	workA(err => {
		if (err) throw err;

		workB(err => {
			if (err) throw err;

			workFoo(err => {
				if (err) throw err;

				res.end();
			});
		});
	});
)

});


app.get('/bar', (req, res) => {
	workA(err => {
		if (err) throw err;

		workB(err => {
			if (err) throw err;

			workBar(err => {
				if (err) throw err;

				res.end();
			});
		});
	});
});
```

haotianw465:
That's a good point. The callback version's 3 subsegments (representing workA, workB and workFoo) are siblings, which is consistent with the promise chain with this PR's patcher code.

I will test the other two examples you mentioned, but I would expect this patcher code to be applied as a general fix for the SDK's native promise support. Will let you know once I make more progress on this.

omsmith:
Just to clarify, in case there's any confusion. This fixes "native Promises", as in Promises provided by V8, created in JavaScript.

Promises created in C++ or by async/await still won't be handled (you need to migrate your SDK away from cls to `async_hook` for that).

haotianw465:
Thank you for the clarification. By general fix I mean the general fix for native promise.
We will look into async/await support as a separate task.

haotianw465:
Hi, for trace hierarchy on example 2 and 3, is your application logic similar to that and you saw the subsegment attached to the wrong trace, or that was your analysis based on how CLS works with native Promise?

I did some load testing locally but haven't reproduce the issue you listed. The issue I actually saw is similar to example 1, where workBar are a direct child of WorkA, instead of attached to another trace.

And again, with the patch code, workBar is a sibling of WorkA, which is the expected behavior.

omsmith:
We have similar logic in our applications.

We can be sure something is wrong, since it's not hard for us to pick out a trace for `GET /ping`, a route which just returns 200 OK (plus or minus middleware), being given subsegments of dozens of dynamodb and SNS calls, something which only occurs on a POST on another route. We've also seen said POSTs to be seemingly missing calls out to dynamo and sns, though we initially figured there may have been some max segment limit.

I have not deployed this patch across the services I manage to see if everything is magically fixed though. We did have a similar problem in the past when using a different APM product which also used CLS. At that time we avoided the specific use of the pattern and saw the problem go away.

Another co-worker noted it fixed their applications while interacting with our TAMs, but re-reading their comments, it was perhaps moreso the nested subsegments being an issue that you just described.

All that said, I _believe_ the test I provided shows the issue occurring rather reliably? Can you think of something else that would be causing it to fail?

omsmith:
I unfortunately haven't tested the examples I provided (I wrote them up directly in GitHub) against X-Ray. Our interaction back and forth here has mostly ended up in my evenings due to other priorities. I do have an instance spun up with my examples in place, but wasn't seeing _anything_ coming through in x-ray, and couldn't spend terribly much more time on fighting with that :smile:.

Apologies for that.

omsmith:
Just wanted to send an update that I got the issues figured out I was having with segments not getting sent to x-ray. Got those fixed up, and had to fix up a few other issues I had with the examples.

Definitely reproduced though. Getting it all compiled in a gist for you!

omsmith:
Fixed examples and screenshots of trace results here: https://github.com/omsmith/aws-xray-sdk-node-pull-11-examples

haotianw465:
Thank you so much for putting time on those write up and the patch code and we really appreciate it. Sorry that I was not able to reproduce example 2 and 3.

Using the example code you provided I am able to reproduce exactly what you saw. Now we know for sure the patch addressed two issues: 

- Nested subsegments on promise chain.
- Wrongly attached subsegments on shared promise.

We will schedule a release with the patch code as soon as possible.

haotianw465:
To update, the interface to apply the patch becomes 
`AWSXRay.capturePromise();`

ashwgupt:
Hello,

That's really a great work here and I think it shall help those trying to wrap their HTTP calls via ES6 promise based clients, such as Axios etc (such as us).

Can you please advise when is this patch schedules to come out for us to use, as we fail to instrument the http calls using XRay at the moment when we use Axios client for the purpose.

Also what shall be need to in conjunction of `AWSXRay.captureHTTPsGlobal(require('http'));` to make it work?

ashwgupt:
Guys I realised that, the patch is already available in the version we use 1.2.0, and simply adding AWSXRay.capturePromise() solved the problem we had with Axios (or any Promise based http client).

Thanks a lot again guys, for solving it :-)

For any user that want to use such a http client, below should be all that needed to capture http globally:

`
const AWSXRay = require('aws-xray-sdk');
AWSXRay.captureHTTPsGlobal(require('http'));
AWSXRay.capturePromise();
const AxiosWithXray = require('axios');
`

PR #17, opened 2018-01-23
Add support for s3 in parameter whitelist file

glicht:
Hi,

Similar to my pull request for [java sdk](https://github.com/aws/aws-xray-sdk-java/pull/9) here is the whitelist file with s3 support for nodejs.

Fields I've included in the request_parameters: Bucket (renamed to bucket_name), Key, VersionId, Prefix, CopySource. 

I've used the following js script to generate the S3 paramters: https://github.com/functionalone/aws-xray-parameter-whitelist-node/blob/master/src/util/generate-param-whitelist.ts 

Best,

Guy

### DIFF
--- a/packages/core/lib/resources/aws_whitelist.json
+++ b/packages/core/lib/resources/aws_whitelist.json
@@ -330,6 +330,619 @@
           ]
         }
       }
+    },
+    "s3": {
+      "operations": {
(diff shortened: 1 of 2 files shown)


haotianw465:
Thank you for the contribution and we really appreciate it.

Issue #20, opened 2018-01-30
Documentation for captureAWS caveat is confusing.


thomasmichaelwallace:
I'm on the cusp of using aws-xray but I have read this caveat for `var AWS = captureAWS(require('aws-sdk'));` about 10 times, and I’m just not getting it:

> Create new clients as usual. Be sure any outgoing calls that are dependent on another async function are wrapped, or duplicate segments might leak.

* If I’m in a lambda function handler and all I do is call into AWS APIs, awaiting the `.promise()` version, do I have to do anything else but wrap the module in `captureAWS`?
* Am I right in understanding _client_ to mean something created using `new AWS.s3()` etc? 
* Isn’t anything I do with a client an outgoing call (e.g. `s3.getObject()`)? Or does “outgoing calls” in this context mean asynchronous actions not using said _clients_ (e.g. `rp.post()`).
* By ‘dependent on’, does it mean that anything in `.then()` (or a babel transpiled await)?

For the sake of people like me, can someone please explain this to me like I'm five (perhaps with an example of when you do, and don't, need to start wrapping things in captureAsyncFunc?).

Assuming I get it, I'm happy to digest it all back into a PR for the documentation to save some confusion for the next person like me!

haotianw465:
Thank you for your feedback and sorry for the confusion. The doc was written before native Promise came out and the SDK doesn't officially support native Promise due to how [CLS](https://github.com/othiym23/node-continuation-local-storage) works. This native promise issue was already addressed here #11 . We will prepare a release soon and add Promise related content to our docs.

For non-Promise related usage you could check this sample app https://github.com/awslabs/aws-xray-rekognition-lambda-sample which uses two Lambda functions and each of them calls some AWS services. Feel free to follow up any question regarding the sample Lambda functions.

You are very welcome to submit a PR on improving the documentation.

thomasmichaelwallace:
OK- so for now using the aws-sdk `.promise()` functions the returned AWSRequest objects don't work? e.g. `s3.getObject().promise()`?

Thanks for the example- I had seen that, but it doesn't really indicate why it's ok to use the asynchronous `cognitoidentity.getCredentialsForIdentity` outside of `captureAsyncFunc` but nothing else?

haotianw465:
I will do some testing around using `.promise()` for AWS call capture and update `README` for what is supported and what is not on latest node version.

[...]

PR #21, opened 2018-02-02
Remove non sequitur mixing segment logic with captureAWS (#20)

thomasmichaelwallace:
After some discussion (see #20) and testing I have removed the commentary about `captureAsyncFunction` usage attached to the the `captureAWS` documentation:

> Create new clients as usual. Be sure any outgoing calls that are dependent on another async function are wrapped, or duplicate segments might leak.

It seems that this is about logical grouping (subsegments) in general, and there's nothing special about the `captureAWS` that function would need you to use these wrappers any differently than you normally would. And, in fact, the clients created from this function all behave perfectly without them (of course, they don't generate their own subsegements, but then, nothing else does).

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -287,11 +287,7 @@ a new context, see: https://github.com/othiym23/node-continuation-local-storage.
 
     var AWS = captureAWS(require('aws-sdk'));
 
-    //Create new clients as usual
-    //Be sure any outgoing calls that are dependent on another async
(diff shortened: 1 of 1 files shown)


haotianw465:
There are other places that have this same comments that need to be removed. And we will add more notes on `captureAsyncFunc` section about logical grouping. But will merge this change. Any further PR on improving doc is welcome.

PR #39, opened 2018-06-01
dont add error for `null` when closing segments

mattcasey:
*Issue #, if available:*

*Description of changes:*

If the value of the first argument to segment.close is `null`, then it throws an exception. This makes it a bit harder to use, as I have to check the errors myself before passing them in after any callback. I am re-using the same check that's in attributes/subsegment.js, where anything that's falsey is ignored.

### DIFF
--- a/packages/core/lib/segments/segment.js
+++ b/packages/core/lib/segments/segment.js
@@ -315,7 +315,7 @@ Segment.prototype.close = function(err, remote) {
   if (!this.end_time)
     this.end_time = SegmentUtils.getCurrentTime();
 
-  if (!_.isUndefined(err))
+  if (err)
     this.addError(err, remote);
 
   delete this.in_progress;


haotianw465:
Hi there, thank you for the feedback.

The current behavior of `close()` will explicitly reject invalid error, which has its own benefit. You can see the full intention here and please let me know if this exception is the one you saw: https://github.com/mattcasey/aws-xray-sdk-node/blob/b09c9d6a993cc17cc79be8887981ac97df522b85/packages/core/lib/segments/segment.js#L215.

I totally understand your use case of the implicit ignore but this expectation might not represent the whole SDK user group. The PR will change the current behavior which might break other users' use cases.

Could you provide a code snippet of how you are manually closing the segment right now so we can take a closer look? Also since `close()` is a public API of `segment` class you can replace it with your own implementation. Please let me know if this is an acceptable solution for you.

haotianw465:
Hi there, any update?

PR #42, opened 2018-06-28
When the persistent socket is disabled, batch segment sends and share a UDP socket for each batch

mdlavin:
When running in a Lambda function and when a segment contains a *lot* of sub-segments, it's possible to exhaust the number of file descriptors and have errors reporting the segments. This change will allow many segments that are sent at the same time to share a single socket. This change means many fewer sockets will be created and should have at least two advantages:

1. Should be more efficient, but I have not measured it
2. It avoids the errors of running out of file handles for large segments. I have verified this by deploying  patched version in our our environment and seeing the fix.

### DIFF
--- a/packages/core/lib/segment_emitter.js
+++ b/packages/core/lib/segment_emitter.js
@@ -8,10 +8,18 @@ var DEFAULT_PORT = 2000;
 var PROTOCOL_HEADER = '{"format": "json", "version": 1}';
 var PROTOCOL_DELIMITER = '\n';
 
-function batchRun (work, callback) {
+/**
+ * Sends a collection of data over a UDP socket. This method
+ * is designed to be used by `atomic-batcher` as a way to share
+ * a single UDP socket for sending multiple data blocks.
+ * 
+ * @param {[object]} ops - Details of the data to send
+ * @param {Function} callback - The function to call when done
+ */
+function batchSendData (ops, callback) {
   var client = dgram.createSocket('udp4');
 
-  processWork(client, work, 0, function () {
+  executeSendData(client, ops, 0, function () {
     try {
       client.close();
     } finally {
@@ -20,17 +28,33 @@ function batchRun (work, callback) {
   });
 }
 
-function processWork (client, work, index, callback) {
-  if (index >= work.length) {
+/**
+ * Execute sending data starting at the specified index and
+ * using the provided client.
+ *  
+ * @param {Socket} client - Socket to send data with
+ * @param {[object]} ops - Details of data to send
+ * @param {number} index - Starting index for sending
+ * @param {Function} callback - Function to call when done
+ */
+function executeSendData (client, ops, index, callback) {
+  if (index >= ops.length) {
     callback();
     return;
   }
 
-  sendMessage(client, work[index], function () {
-    processWork(client, work, index+1, callback);
+  sendMessage(client, ops[index], function () {
+    executeSendData(client, ops, index+1, callback);
   });
 }
 
+/**
+ * Send a single message over a UDP socket.
+ * 
+ * @param {Socket} client - Socket to send data with
+ * @param {object} data - Details of the data to send
+ * @param {Function} batchCallback - Function to call when done
+ */
 function sendMessage (client, data, batchCallback) {
   var msg = data.msg;
   var offset = data.offset;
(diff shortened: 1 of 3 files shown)


mdlavin:
@haotianw465 or @awssandra any thoughts on this? We're using a patched version of this library in our deployments right now and I'd love to move to an official version if this can be merged. If there are any changes you'd like to see I'm happy to do the work.

packages/core/lib/segment_emitter.js
```diff
@@ -1,12 +1,70 @@
...
+  var work = {
+    msg: msg,
+    offset: offset,
+    length: length,
+    port: port,
+    address: address,
+    callback: callback
+  };
+
+  this.sendBatcher(work);
```
haotianw465: should this be renamed batchSend?

packages/core/lib/segment_emitter.js
```diff
@@ -1,12 +1,70 @@
...
 
+var batcher = require('atomic-batcher');
 var logger = require('./logger');
 
 var DEFAULT_ADDRESS = '127.0.0.1';
 var DEFAULT_PORT = 2000;
 var PROTOCOL_HEADER = '{"format": "json", "version": 1}';
 var PROTOCOL_DELIMITER = '\n';
 
+function batchRun (work, callback) {
```
haotianw465: You have a variable named `work` defined in line46 which is a dictionary contains ip and port etc. Is this work the same here? Should it be the actual `ops`?

packages/core/lib/segment_emitter.js
```diff
@@ -1,12 +1,70 @@
...
+}
+
+function processWork (client, work, index, callback) {
+  if (index >= work.length) {
+    callback();
+    return;
+  }
+
+  sendMessage(client, work[index], function () {
+    processWork(client, work, index+1, callback);
```
haotianw465: In the usage guide https://www.npmjs.com/package/atomic-batcher it seems like the batcher automatically batch the ops? What is the intention to do recursive calls with `processWork` and how does the batcher decide the proper batcher size?

packages/core/lib/segment_emitter.js
mdlavin: Yes, that's a much better name. I'll change it.

packages/core/lib/segment_emitter.js
mdlavin: Good catch! The name `ops` sounds good to me.

packages/core/lib/segment_emitter.js
mdlavin: Yes, the `atomic-batcher` does batch the ops. The default batch size is as many as big as possible since the last batch. It doesn't introduce any delay in sending, it just queues up the work while existing work is in progress.

mdlavin: I used recursion in `processWork` only because it was an easy way to iterate over the work using callbacks. I thought about bringing in a librayr like `async` to use `async.series` but that felt like too much. Do you have a suggestion that you prefer?

mdlavin:
@haotianw465 checkout my new updates. I've improved the function and variable names. I also added comments to the new functions that explains their use

haotianw465:
I have two concerns. One is that this might cause stack overflow. Based on my understanding `the number of recursive calls == ops.length` per batch execution but the batch size is decided by the library and likely to be aggressive based on what you said "The default batch size is as many as big as possible". But considering the load you have would otherwise exhaust the file descriptors and your application is running fine with batching, I think this should not be an issue.

The other thing is that the socket send buffer for each socket has a fixed size per OS. Although now the  emitter uses less sockets, the requirement for a single socket throughput is higher since it is sending multiple segment documents to a single socket and the send buffer needs to be cleared fast enough (which means the X-Ray daemon needs to read fast enough) so the packet will not be dropped (or send blocked depending on OS config) due to buffer full. The UDP traffic is local so this is likely not an issue neither, but I don't have data or evidence to say there is no chance of such case happening.

Having said that, I think this is a more optimized strategy for most use cases. We might have to provide an opt-in option to switch back to one socket per send for some special cases. I'm happy to merge this but I would like to hear from your thoughts on these concerns. It would also be good if you increase the subsegment size of your Lambda function and test it out. The simplest way is to add a big chunk of metadata to each subsegment. This test makes sure the PR still works best for you when the subsegments grow into a certain size as you add more information to it.

mdlavin:
Re the stack overflow. I don't think it's an issue because the callback should be called asynchronously. Each call to sendMessage is an IO action and so it should generate a new IO event and callback when it's done. If you are still concerned, setImmediate could be used to avoid the issue. I don't think it's needed.

Re socket send buffer. This code moves the Lambda mode slightly towards the performance of the non-Lambda mode. So if there are no issues with the fully used socket I don't expect there will be any issues with this version either. It should be less efficient than the non-Lambda mode.

I think I understand both of your concerns, but I don't think that either one needs a change. We have been running with this change and some pretty large X-Ray segment lists without any trouble in their scenario. Like you said, before this it was a crash instead =)

haotianw465:
Done. This PR is released in 1.3.0 with some other changes. https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/CHANGELOG.md

PR #55, opened 2018-09-21
Move aws-sdk out of the dependencies list

### LINKED ISSUE #54: aws-xray-sdk-core bloats package bundles
When deploying on AWS Lambda, `aws-sdk` does not need to be bundled since it is already present in the Node.js runtime. This yields significant bundle savings, but... not if you're using `aws-xray-sdk-core`.

The reason for this is `aws-xray-sdk-core` requires `aws-sdk` as a normal dependency. It should be a `peerDependency` and possibly a `devDependency`.

compwright:
*Issue #, if available:* Resolves #54

*Description of changes:* Moves `aws-sdk` out of the `aws-xray-sdk-core` package dependencies, so as to not bloat project bundles that will be deployed to AWS Lambda Node.js runtime.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -15,7 +15,6 @@
   },
   "dependencies": {
     "atomic-batcher": "^1.0.2",
-    "aws-sdk": "^2.304.0",
     "continuation-local-storage": "^3.2.0",
     "date-fns": "^1.29.0",
     "lodash": "^4.17.10",
@@ -24,6 +23,7 @@
     "winston": "^2.4.4"
   },
   "devDependencies": {
+    "aws-sdk": "^2.304.0",
     "chai": "^3.5.0",
     "eslint": "^3.10.2",
     "grunt": "^1.0.3",
@@ -33,6 +33,9 @@
     "nock": "^8.0.0",
     "sinon": "^4.0.0",
     "sinon-chai": "^2.8.0"
+  },
+  "peerDependencies": {
+    "aws-sdk": "^2.304.0"
   },
   "scripts": {
     "test": "./node_modules/.bin/mocha --recursive ./test/ -R spec"


jplock:
It would be fantastic if someone at AWS could review this PR and cut a new release.

haotianw465:
This is definitely a reasonable request and we will work with you to fix this as soon as possible. Unfortunately we can't simply accept and merge this PR.

In version `2.0` the SDK has two timers that use `aws-sdk` to call X-Ray service for fetching sampling rules if running on non-Lambda: https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/middleware/sampling/service_connector.js. In the current implementation this is not an issue when the user code patches `aws_sdk` as the two timers use different copies with the user code. But with changing to `peerDependencies` the X-Ray SDK could use the patched `aws_sdk` while for the timers there is obviously no tracing context and those calls should not be captured. The `aws_sdk` patcher code also needs to be fixed so this change doesn't break v`2.0` on non-Lambda workload.

The other concern is that for customers running on non-Lambda workload with npm 3 or later, this could be a breaking change since starting with `npm v3` it will not install peerDependencies if a package is not explicitly declared as a dependency on higher dependency tree. Developers also would need to manually resolve the conflict if there is one. See https://github.com/npm/npm/issues/6565. (A good thing but still a breaking behavior.)

I agree this PR is a low hanging fruit for Lambda customers but we want to make sure we are also doing the right fix for our non-Lambda customers. Please correct me if I misunderstood any implication of this PR or if you have any thoughts regarding the concerns.

jplock:
@haotianw465 is there a way to split X-ray core into separate lambda and non-lambda packages to optimize for those usage models?

compwright:
@haotianw465 would it work to do this in lib/middleware/sampling/service_connector.js:

```javascript
var XRay = require('aws-sdk/clients/xray');
```

...

```javascript
var ServiceConnector = {
  ...
  client: new XRay({endpoint: util.format('http://%s:%d', DaemonConfig.tcp_ip, DaemonConfig.tcp_port)}),
  ...
}
```

The benefit here is that at least you could tree-shake away the rest of the duplicate aws-sdk package with something like https://github.com/serverless-heaven/serverless-webpack.

An even better solution (that requires more work) would be to update the aws-sdk project with lerna and publish each service client as its own package, similar to the way the aws-xray-sdk project is structured and published. Then you could just require the XRay client and we would all be on our merry ways. But that's a subject for another repo on another day.

haotianw465:
@jplock Yes we will look into separating it into two packages. Right now the two use cases (Lambda and non-Lambda) are tightly coupled so this work is not trivia.

@compwright Thank you for your suggestion. We will look into the repo link you provided. We've also forwarded your request on modularizing AWS Nodejs SDK to AWS SDK team.

Right now we are still maintaining 1.x for bug fixes and the only new feature for X-Ray SDK 2.x is service sampling rules which is not applicable on AWS Lambda. You can continue using 1.x if the 2.x bundle size is an issue. We will update this issue as soon as we make progress.

jafl:
If aws-xray-sdk was provided by default, the way aws-sdk is, then this wouldn't be an issue.

dougmoscrop:
You can omit the aws-sdk during Lambda packaging using most toolsets, but bundling Lodash and Winston is surely not helping. But also, yeah, provided by default!

haotianw465:
The size issue was resolved on webpack case https://github.com/aws/aws-xray-sdk-node/pull/73. AWS Node SDK just released v3 in preview which is modularized https://aws.amazon.com/blogs/developer/new-aws-sdk-for-javascript-developer-preview/. I'm closing this PR in favor of https://github.com/aws/aws-xray-sdk-node/issues/82. To @jafl's point we are working with Lambda team with what you are suggesting.

jplock:
Thanks @haotianw465!  Will you be releasing a new version of this library with the fixes from #73 in it prior to implementing #82?

dougmoscrop:
Not everyone uses webpack, etc.

https://github.com/aws/aws-xray-sdk-node/search?q=lodash&unscoped_q=lodash

I think Lodash could be replaced with vanilla JS in many cases, and at least import the lodash.whatever methods for the ones that can't, not the whole package.

Logging - pino is quite small and fast

haotianw465:
@jplock yes we will prepare a release soon regardless of progress on #82. 

@dougmoscrop I created #82 so we can fix the non webpack use case and non-lambda use case. We will have to wait for AWS SDK v3 becomes GA before directly depend on it. For loadash changes feel free to open a separate issue or PR and we will keep improving the bundle size.

PR #57, opened 2018-10-17
Call getLogger()

moonthug:
*Issue #, if available:*

*Description of changes:*
Fix TypeError where `getLogger.error` is undefined. It should call `getLogger` and then call `error`


haotianw465:
Thank you for your contribution. We will release the fix soon.

Issue #60, opened 2018-10-19
Tracing Context Preservation (async/await and various of libraries)


haotianw465:
## Goal
The goal of this issue is to track all tracing context preservation related problems, dive deep on why some of them work and some of them don't, discuss and gather feedback for a way going forward to provide much better customer experience with X-Ray node tracing and further contribute/shape the node context propagation in general.

## Problem Statement
The Node.js event loop model has its downsides. One of those is the context preservation through the call chain within a single http request. It is hard to bind arbitrary data (user name, transaction id, tracing context etc) to a full request/response cycle and all relevent context stops at event loop. It's hard to know what is the context for the next callback being executed by event loop.

There is a library [continuation-local-storage](https://github.com/othiym23/node-continuation-local-storage) (CLS for short) that solved most of the problem by binding request context to callbacks.   However, this library has two limitations:
* It's both a bug and a feature that CLS doesn't work with all node modules. The full reasoning from CLS author is [here](https://github.com/othiym23/node-continuation-local-storage/issues/6#issuecomment-24114844). In short it's difficult for CLS to know whether it should bind the context or not all the time.
* Node community evolves fast and with the major adoption of async/await syntax from Node 8 CLS doesn't fit some new use cases anymore.

## Challenges for X-Ray SDK for Node.js
With all the context preservation issues listed above, they are solvable for most of the node application owners who need to preserve contextual data within requests.
* For the first issue, CLS provides various of wrappers with name like `cls-*` for explicitly binding context to callbacks in a particular module. [Here](https://github.com/othiym23/cls-redis) is an example fork for `node-redis` that patches Redis driver for context preservation.
* For the second issue (async/await), [cls-hooked](https://github.com/Jeff-Lewis/cls-hooked) are reported to be worked well with Node 8 so far by several SDK users. [Zone.js](https://github.com/angular/zone.js/) is also a potential candidate for the new dependency of the SDK for context binding.

These workarounds are not ideal but work for each service owner as each service only has limited dependency and the owner knows what dependencies the service depends on. It's inevitable and necessary work for service owners to understand this problem space and make necessary code changes since it is business data (user name, transaction id etc) through a request/response cycle.

[...]

PR #74, opened 2018-11-16
feat: switches continuation-local-storage to cls-hooked for async support

chrisradek:
*Issue #, if available:*
#60 
*Description of changes:*
Switch `continuation-local-storage` dependency to `cls-hooked`.

I tested against the 3 scenarios listed out in #11.
First, I tested using just promises. With the updated dependency, I no longer needed to call `xray.capturePromise()` to get the correct hierarchy of subsegments to segments. Continuing to call `capturePromise()` also does not change the the segment hierarchy.

Secondly, I update my express routes to use async functions and await https requests. This resulted in the same segment hierarchy as the promises example.

I have not looked too far into the concern pointed out in #60 with using cls-hooked, namely: https://github.com/Jeff-Lewis/cls-hooked/issues/23

This probably warrants some more investigation into how cls-* libraries get around some of these issues, and if they are truly unique to cls-hooked.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,14 @@ node_js:
 
 sudo: false
 
+matrix:
+  include:
+    - node_js: "8"
+      env: TEST_ASYNC=test-async
+    - node_js: "10"
+      env: TEST_ASYNC=test-async
+
 script:
   - lerna bootstrap --hoist
   - lerna run test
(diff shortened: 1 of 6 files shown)


chrisradek:
Reset the base to the `experimental-hooks` branch.
My main concern is that by moving to `cls-hooks` from `continuation-local-storage`, users relying on modules like `cls-bluebird` to add continuation-local-storage support to their dependencies will surprisingly stop generating traces.

This branch gives us a place to test the change and begin to see the impact it has.

PR #81, opened 2018-12-12
fix: handle null/undefined as last query value

M-TGH:
*Description of changes:*

Was beforehand throwing a TypeError when trying to access the constructor property on `null` or `undefined`:

```js
connection.query('SELECT a FROM b', null);
// without xray --> resolves query
// with xray    --> throws TypeError: Cannot read property 'constructor' of null  
```

As far as I know null and undefined are the only data types that will encounter this problem, hence a loose null check should suffice.

### DIFF
--- a/packages/mysql/lib/mysql_p.js
+++ b/packages/mysql/lib/mysql_p.js
@@ -92,7 +92,7 @@ function resolveArguments(argsObj) {
       args.callback = typeof argsObj[1] === 'function' ? argsObj[1] : (typeof argsObj[2] === 'function' ? argsObj[2] : undefined);
     }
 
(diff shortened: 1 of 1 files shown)


haotianw465:
Thank you for reporting the issue. I believe the postgres package has exactly the same issue. https://github.com/aws/aws-xray-sdk-node/blob/master/packages/postgres/lib/postgres_p.js#L47. 

Could you add a test for mysql? If you have extra time would you mind making this change to postgres patcher as well? If not we will merge this and fix postgres using the same code you provide. Please let me know.

M-TGH:
Wasn't too sure what way to test it as it's a bit of an edge case, but these test cases will fail without the patch.

Also added the patch to the postgres patcher as it did indeed have the same behaviour.

Let me know if there's anything else.

Issue #88, opened 2019-01-02
Remove or optimize engines section


EugeneDraitsev:
If you will try to run `yarn` with node version >11, f.e. inside docker image node:latest you'll get an error:

```
error aws-xray-sdk@2.1.0: The engine "node" is incompatible with this module. Expected version ">= 4.x <= 10.x". Got "11.6.0"
```
It appears because of engines section in your package.json
```
  "engines": {
    "node": ">= 4.x <= 10.x",
    "npm": ">= 2.x <= 5.x"
}
```

So, question is do you really need node >= 4.x <= 10.x, or maybe you should replace it with >= 4.x and allow users to use aws-xray-sdk-node with latest version of node?

haotianw465:
Hi @EugeneDraitsev, the SDK is only tested under <=10.x and we do have some work need to be done for testing against v11. I'm OK with removing this restriction and have some wording on the `README` about the version we officially support. I will post some update regarding this change as soon as possible.

PR #89, opened 2019-01-07
chore: removes upper bound on supported node versions

chrisradek:
*Issue #, if available:*
#88 

*Description of changes:*
Removes the upper bound Node.js version restriction from all packages. Also adds the latest stable version of node.js (currently 11.x) in the travis build matrix so tests are run against the latest version of node.


chrisradek:
I've tested it with Node 11, but I can add something that says it is not fully tested on versions past 11.

I can also set up travis to run daily so the unit tests at least get run against the latest version of node travis has available.

Issue #87, opened 2018-12-30
[mysql] Support for getConnection of a pool


otbe:
Hi there,

I wonder why I dont see any mysql data in my xray traces and therefore I read over the source code of the `captureMysql`function. If I get it right it only patches `execute` and `query` on the return objects of `createConnection` and `createPool`. 
I use [typeorm](https://github.com/typeorm/typeorm) as my ORM tool and it uses under the hood [`getConnection`](https://github.com/typeorm/typeorm/blob/master/src/driver/mysql/MysqlDriver.ts#L627) of a pool. 

Any chances to get support for this? Maybe I can help?

Thanks!

BTW: I had to switch to `2.1.0-experimental.1` to get xray working at all. Maybe its related to `cls-hooked` which I use on my own.

bradleyharrisonboat:
@otbe I am in the exact same position as you are with the typeORM getConnection. Have you had any luck?  This is important enough that I may fork but I loathe doing that if it cannot be added back into the main stack

chrisradek:
@otbe 
Thank you for the request. We can take a look at supporting `getConnection`, and would love any help in supporting it, even if that just means providing a simple test case we can verify against.

Side note: Are you using async/await in your code? The experimental tag of the SDK does use cls-hooked instead of continuation-local-storage, with one benefit being it supports async/await.

otbe:
@chrisradek I use async/await in my code but TS transpiles it down to generators because some issues with another lib. If I use the latest version of this lib I get a strange error from [this](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/context_utils.js#L61) line. It seems that `getNamespace` returns null/undefined in my situation :/

PR #98, opened 2019-02-11
feat: patch connections created by a pool's getConnection

M-TGH:
*Issue #, if available:* #87

*Description of changes:*

Adds a simple patchObject call around connections returned from a pools `getConnection`.
Far as I could tell mysql doesn't take any other arguments for this, so it assumes the first argument is the callback.

I'm unsure what mysql2 makes of this, but I assume they promisified it which I didn't yet support. Would be nice to hear from someone who uses mysql2 what changes are still needed to support mysql2. I'll be looking into adding it but would like confirmation on what parameters it takes and what it returns.

### DIFF
--- a/packages/mysql/lib/mysql_p.js
+++ b/packages/mysql/lib/mysql_p.js
@@ -66,6 +66,25 @@ function patchCreatePool(mysql) {
   };
 }
 
+function patchGetConnection(pool) {
+  var baseFcn = '__getConnection';
+  pool[baseFcn] = pool['getConnection'];
+
+  pool['getConnection'] = function patchedGetConnection() {
+    var args = arguments;
+    var callback = args[0];
+
+    if(callback instanceof Function){
+      args[0] = (err, connection) => {
+        if(connection) patchObject(connection);
+        return callback(err, connection);
+      }
+    }
+
+    return pool[baseFcn].apply(pool, args);
+  }
+}
+
 function patchObject(connection) {
   if (connection.query instanceof Function) {
     connection.__query = connection.query;
@@ -75,6 +94,10 @@ function patchObject(connection) {
   if (connection.execute instanceof Function) {
     connection.__execute = connection.execute;
     connection.execute = captureOperation('execute');
+  }
+
+  if(connection.getConnection instanceof Function){
+    patchGetConnection(connection);
   }
 }
 
--- a/packages/mysql/lib/mysql_p.js
+++ b/packages/mysql/lib/mysql_p.js
@@ -26,6 +26,7 @@ module.exports = function captureMySQL(mysql) {
 
   patchCreateConnection(mysql);
   patchCreatePool(mysql);
+  patchCreatePoolCluster(mysql);
 
   return mysql;
 };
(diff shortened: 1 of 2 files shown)


M-TGH:
Added what I think should be all that's needed for mysql2 support, still wouldn't mind if someone who's more familiar with mysql2 could take a look at it as well.

darcoli:
Tried using this within our application which uses `typeORM` with the `mysql` driver, however I kept getting the following exception:
```
'RangeError: Maximum call stack size exceeded',
'    at Subsegment.init (/var/task/dist/webpack:/application/node_modules/aws-xray-sdk-core/lib/segments/attributes/subsegment.js:32:1)',
'    at new Subsegment (/var/task/dist/webpack:/application/node_modules/aws-xray-sdk-core/lib/segments/attributes/subsegment.js:25:1)',
'    at Segment.addNewSubsegment (/var/task/dist/webpack:/application/node_modules/aws-xray-sdk-core/lib/segments/segment.js:179:1)',
'    at PoolConnection.__query (/var/task/dist/webpack:/application/node_modules/aws-xray-sdk-mysql/lib/mysql_p.js:152:1)',
'    at PoolConnection.__query (/var/task/dist/webpack:/application/node_modules/aws-xray-sdk-mysql/lib/mysql_p.js:178:1)',
'    at PoolConnection.__query (/var/task/dist/webpack:/application/node_modules/aws-xray-sdk-mysql/lib/mysql_p.js:178:1)',
'    at PoolConnection.__query (/var/task/dist/webpack:/application/node_modules/aws-xray-sdk-mysql/lib/mysql_p.js:178:1)',
'    at PoolConnection.__query (/var/task/dist/webpack:/application/node_modules/aws-xray-sdk-mysql/lib/mysql_p.js:178:1)',
'    at PoolConnection.__query (/var/task/dist/webpack:/application/node_modules/aws-xray-sdk-mysql/lib/mysql_p.js:178:1)' ] } } }   
```

I did get it to work by adding further conditions to avoid the infinite recursion - changing the function `patchObject()` as follows:
```
function patchObject(connection) {
  if (connection.query instanceof Function) {
    if (!connection.__query) {
      connection.__query = connection.query;
      connection.query = captureOperation('query');
    }
  } 
    
  if (connection.execute instanceof Function) {
    if (!connection.__execute) {
      connection.__execute = connection.execute;
      connection.execute = captureOperation('execute');
    }
  }
      
  if (connection.getConnection instanceof Function) { 
    if (!(connection['__getConnection']) ) {
      patchGetConnection(connection);
    }
  }
  return connection;
}
```

M-TGH:
Thanks @darcoli, I've added your feedback in e0fadd4.

darcoli:
@M-TGH thanks for adding the changes - did not get any issues so far ;)

Any chance of getting this merged any time soon?

packages/mysql/lib/mysql_p.js
```diff
@@ -66,16 +66,42 @@ function patchCreatePool(mysql) {
   };
 }
 
+function patchGetConnection(pool) {
```
chrisradek: Would this also work with `PoolCluster` if instead of grabbing the callback from the 1st argument, you grabbed it from the last argument?

https://github.com/mysqljs/mysql#poolcluster

M-TGH: Looks like it would work, I'll add it into this PR, if you prefer me to split it up into a separate PR let me know (it may make the PR a bit big in size).

M-TGH:
@chrisradek I've added basic support for mysql.createPoolCluster in dfc1c55 (getConnection & query). Let me know if you would prefer it in a separate PR.

I've also added support for poolCluster.of in db8e782, however I'm on the fence about patching of this way since it's rather ambiguous.

M-TGH:
Good catch on the args, added a fix in 5bb11c8.

packages/mysql/lib/mysql_p.js
```diff
@@ -66,16 +67,72 @@ function patchCreatePool(mysql) {
...
+
+    if (callback instanceof Function) {
+      args[args.length-1] = (err, connection) => {
+        if(connection) patchObject(connection);
+        return callback(err, connection);
+      }
+    }
+
+    var result = pool[baseFcn].apply(pool, args);
+    if (result instanceof Promise) return result.then(patchObject);
```
chrisradek: Sorry I missed this before, but could you instead check if `result` has a method `then`? I've seen this check bite us in other libraries when a user has replaced the native `Promise` class with a different once, e.g. bluebird.

M-TGH:
Updated the check in 1937797.
Whilst replacing this `instanceof Promise` check I noticed there's 2 more:
https://github.com/aws/aws-xray-sdk-node/blob/a8741c0151fbc43df621afd50229616533687d80/packages/mysql/lib/mysql_p.js#L39
https://github.com/aws/aws-xray-sdk-node/blob/a8741c0151fbc43df621afd50229616533687d80/packages/mysql/lib/mysql_p.js#L57

I'll replace those in a separate PR since they weren't added in this one, however let me know if you prefer it included in this PR.

Issue #84, opened 2018-12-19
Use lodash modules for package dependencies


chris-feist:
To reduce the bundle size, lodash method modules should be used for package dependencies instead of the whole lodash library. See https://www.npmjs.com/search?q=keywords:lodash-modularized.

Looking at #43, it appears there are only a handful that need to be added.

chrisradek:
@chris-feist 
Thanks for the suggestion. We're happy to consider safe changes that can reduce the size of the SDK. It also looks like we don't have to use lodash for everything we're using it for today, and could possibly just use vanilla JS instead in most cases.

Just curious, when you mention reducing the bundle size, do you mean when using a tool like webpack to create a bundle, or the contents of the node_modules directory?

PR #99, opened 2019-02-11
refactor: remove lodash from runtime dependencies

M-TGH:
*Issue #, if available:* #84

*Description of changes:*

Removes use of lodash functions making lodash no longer required on runtime.
May have a slight regression in performance on the replacement of omit. 
Would prefer using destructuring but that's not supported below node v6.

### DIFF
--- a/packages/core/lib/segments/attributes/subsegment.js
+++ b/packages/core/lib/segments/attributes/subsegment.js
@@ -116,7 +116,7 @@ Subsegment.prototype.addPrecursorId = function(id) {
  */
 
 Subsegment.prototype.addAnnotation = function(key, value) {
-  if (!['boolean', 'number', 'string'].includes(typeof value)) {
+  if (!(typeof value === 'boolean' || typeof value === 'string' || (typeof value === 'number' && isFinite(value)))) {
     throw new Error('Failed to add annotation key: ' + key + ' value: ' + value + ' to subsegment ' +
       this.name + '. Value must be of type string, number or boolean.');
   } else if (typeof key !== 'string') {
--- a/packages/core/lib/segments/attributes/subsegment.js
+++ b/packages/core/lib/segments/attributes/subsegment.js
@@ -1,11 +1,3 @@
-var omit = require('lodash/omit');
-var isEmpty = require('lodash/isEmpty');
-var isBoolean = require('lodash/isBoolean');
-var isFinite = require('lodash/isFinite');
-var isString = require('lodash/isString');
-var isUndefined = require('lodash/isUndefined');
-var isObject = require('lodash/isObject');
-
 var crypto = require('crypto');
 
 var CapturedException = require('./captured_exception');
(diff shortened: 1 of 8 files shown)


packages/core/lib/segments/attributes/subsegment.js
```diff
@@ -124,15 +116,15 @@ Subsegment.prototype.addPrecursorId = function(id) {
  */
 
 Subsegment.prototype.addAnnotation = function(key, value) {
-  if (!(isBoolean(value) || isString(value) || isFinite(value))) {
+  if (!['boolean', 'number', 'string'].includes(typeof value)) {
```
chrisradek: I like how you've kept the conditional compact, but if the type is a number we also want to make sure it is finite (not Infinity or NaN) using the native `isFinite` function. Something like this:

```suggestion
  if (!(typeof value === 'boolean' || typeof value === 'string' || (typeof value === 'number' && isFinite(value))) {
```

packages/core/lib/segments/attributes/subsegment.js
```diff
@@ -379,12 +371,16 @@ Subsegment.prototype.toString = function toString() {
...
+  var thisCopy = Object.assign({}, this);
+
+  if ('segment' in thisCopy) delete thisCopy.segment;
+  if ('parent' in thisCopy) delete thisCopy.parent;
+  if ('counter' in thisCopy) delete thisCopy.counter;
 
-  if (isEmpty(this.subsegments))
-    ignore.push('subsegments');
+  if (Array.isArray(this.subsegments) && !this.subsegments.length && 'subsegments' in thisCopy)
+    delete thisCopy.subsegments;
```
chrisradek: Nit: Since the body of the if statement is on a separate line, can you wrap this in brackets? That'll make it more clear what our intent is.

packages/core/lib/segments/segment.js
```diff
@@ -346,18 +338,26 @@ Segment.prototype.close = function(err, remote) {
...
+  var thisCopy = Object.assign({}, this);
+
+  if ('counter' in thisCopy) delete thisCopy.counter;
+  if ('notTraced' in thisCopy) delete thisCopy.notTraced;
+  if ('exception' in thisCopy) delete thisCopy.exception;
 
-  if (isEmpty(this.subsegments))
-    ignore.push('subsegments');
+  if (Array.isArray(this.subsegments) && !this.subsegments.length && 'subsegments' in thisCopy)
+    delete thisCopy.subsegments;
```
chrisradek: Nit: Same as above. Since the body of the if statement is on a separate line, can you wrap this in brackets? That'll make it more clear what our intent is.

M-TGH:
Thanks for the feedback, I've added the brackets to the long if statements in 194a023.
Added isFinite in 2f11c21 but was missing a closing bracket, fixed in 0816dd0.

chrisradek:
@M-TGH 
Thanks for making the changes and for your contribution! Merging!

PR #102, opened 2019-02-27
bugfix(core): prevents xray calls made for centralized sampling from being traced

### LINKED ISSUE #100: Failed tracing errors on xray.getSamplingTargets and x-ray.getSamplingRules
I get failed tracing errors on `xray.getSamplingTargets` and `xray.getSamplingRules` calls after enabling tracing for all Amazon clients: 

```const AWS = AWSXRay.captureAWS(require('aws-sdk'))```

The failed traces are polluting the logs, but otherwise, they're not a big issue. I can't, however, figure out how to avoid them:

> "Call xray.getSamplingRules is missing the sub/segment context for automatic mode. Ignoring."
> "Call xray.getSamplingTargets is missing the sub/segment context for automatic mode. Ignoring."

Seems like X-Ray traces its own calls when capturing the entire AWS SDK. I

n my app, I'm actually interested in capturing DynamoDB calls only. I've tried capturing it like this but I don't seem to get any traces.

```
const dynamo = AWSXRay.captureAWSClient(new AWS.DynamoDB({...})
                                            
db = new AWS.DynamoDB.DocumentClient({
      service: dynamo,
      params: ...})
```

Is it maybe because I'm using `AWS.DynamoDB.DocumentClient`?

chrisradek:
*Issue #, if available:*
Fixes #100 

*Description of changes:*
The X-Ray SDK now checks if the AWS SDK client it is instrumenting is the one used internally to poll the x-ray daemon. This prevents traces being generated (and resulting in missing context errors) for x-ray calls made by this SDK.

### DIFF
--- a/packages/core/lib/patchers/aws_p.js
+++ b/packages/core/lib/patchers/aws_p.js
@@ -10,6 +10,7 @@ var contextUtils = require('../context_utils');
 var Utils = require('../utils');
 
 var logger = require('../logger');
+var ServiceConnector = require('../middleware/sampling/service_connector');
 
 var minVersion = '2.7.15';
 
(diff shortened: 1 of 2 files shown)


fjsousa:
Thanks! This should clean some of our logs! Any idea when this is being released?

PR #108, opened 2019-03-29
Avoid logging the error message twice and comply to winston v3 signature

ifroz:
Injecting a winston v3 logger breaks the way how the log is outputted, given that they
do not handle instances of the Error object anymore and the logger fn signature changed

*Description of changes:*

Only log stack trace, not the error object itself in order to comply with both v2 and v3 of winston. 


chrisradek:
@ifroz 
Thank you for your contribution. Merging this in!

PR #116, opened 2019-04-19
chore: sets aws-xray-sdk-core as a dev dependency for restify

chrisradek:
*Description of changes:*
Previously, aws-xray-sdk-restify depended on the version of `aws-xray-sdk-core` that was installed by the root `package.json`, which would cause a different version of the core SDK to be used than what is specified in `peerDependencies` in some cases. Adding the dependency to `devDependencies` as well forces tests to be run against the version we expect.

### DIFF
--- a/packages/restify/package.json
+++ b/packages/restify/package.json
@@ -17,6 +17,7 @@
     "aws-xray-sdk-core": "^2.3.0"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


chrisradek:
I should note that this is what the mysql and postgres x-ray libraries do today.

PR #117, opened 2019-04-19
Revert "refactor: remove lodash from runtime dependencies"

chrisradek:
Reverts aws/aws-xray-sdk-node#99

### DIFF
--- a/packages/core/lib/patchers/http_p.js
+++ b/packages/core/lib/patchers/http_p.js
@@ -6,6 +6,7 @@
  * This module patches the HTTP and HTTPS node built-in libraries and returns a copy of the module with tracing enabled.
  */
 
+var omit = require('lodash/omit');
 var url = require('url');
 
 var contextUtils = require('../context_utils');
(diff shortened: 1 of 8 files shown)


M-TGH:
@chrisradek I'm really sorry these changes were causing issues, do you happen to have a specific error or stack trace so I can look into patching and reopening a new PR? (And adding tests where possible to make sure this doesn't happen again)

chrisradek:
@M-TGH 
Don't worry, our tests were passing but it showed up later on, we missed it too!

The issue that was happening was that when SegmentEmitter.send was called, the Segment that was passed in was missing the `format` method. This was happening because `Object.assign` doesn't copy over methods that exist on an object's prototype, so when `segment.format` was called, the method `format` didn't exist. Specifically, we were seeing this when one of the restify tests were exercising a use-case where a segment had `flush` called. We only caught it due to the fact that our restify tests were actually pulling in 2 versions of the SDK, so only one version had methods stubbed that was masking this issue.

PR #119, opened 2019-04-22
chore: prepares 2.3.1-experimental release

chrisradek:
*Description of changes:*
Prepares release of version 2.3.1-experimental SDK.

Merges in changes from the stable 2.3.1 SDK.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "aws-xray-sdk-core",
-  "version": "2.2.0-experimental.1",
+  "version": "2.3.1-experimental.1",
   "description": "AWS X-Ray SDK for Javascript",
   "author": "Amazon Web Services",
   "contributors": [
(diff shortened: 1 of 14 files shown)


packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "aws-xray-sdk-core",
-  "version": "2.2.0-experimental.1",
+  "version": "2.3.1-experimental.1",
```
luluzhao: Does the package version independent from the version listed in CHANGELOG? Just curious about it since I saw the version listed in changelog is different from this one.

chrisradek: In this case you can see this as version 2.3.1, just with the experimental tag. I haven't been marking the changelog versions as experimental on this branch for the past few releases, but I can if you think that'll be more clear.

luluzhao: That's fine, just want to make sure we don't mess up with the version since I am not familiar with how experimental works. Thank you for your explanation. :)

PR #114, opened 2019-04-18
fix: replace instanceof Promise checks with thennable checks

M-TGH:
*Issue #, if available:* No issue, came up in https://github.com/aws/aws-xray-sdk-node/pull/98#discussion_r276745026.

*Description of changes:* Replaces 2 `instanceof Promise` checks to instead check if .then is a function.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

If preferred I can merge this into the feat/support-pool-get-connection branch.

Wasn't too sure on a way to add tests for this, so refrained on adding them for now. Let me know if you have any ideas.

PR #120, opened 2019-04-22
fix: no longer error when checking .then on variables

M-TGH:
*Issue #, if available:* https://github.com/aws/aws-xray-sdk-node/pull/114#issuecomment-485554819

*Description of changes:*

Prevents errors from occurring on the case that a variable is null or undefined and being checked on whether it's a promise through accessing the .then property.
I think properties are only inaccessible on null and undefined, so as 
far as I know this should fully catch this case. Let me know if there's 
more.

Sorry for the double if you already fixed it, noticed https://github.com/aws/aws-xray-sdk-node/pull/114#issuecomment-485555958 after I already made these changes so I thought I might as well open a PR.

### DIFF
--- a/packages/mysql/lib/mysql_p.js
+++ b/packages/mysql/lib/mysql_p.js
@@ -37,7 +37,7 @@ function patchCreateConnection(mysql) {
 
   mysql['createConnection'] = function patchedCreateConnection() {
     var connection = mysql[baseFcn].apply(connection, arguments);
-    if (connection != null && connection.then instanceof Function) {
+    if (connection && connection.then instanceof Function) {
       connection = connection.then((result) => {
         patchObject(result.connection);
         return result;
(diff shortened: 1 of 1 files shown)


packages/mysql/lib/mysql_p.js
```diff
@@ -37,7 +37,7 @@ function patchCreateConnection(mysql) {
 
   mysql['createConnection'] = function patchedCreateConnection() {
     var connection = mysql[baseFcn].apply(connection, arguments);
-    if (connection.then instanceof Function) {
+    if (connection != null && connection.then instanceof Function) {
```
chrisradek: I think just a truthy check would suffice for all of these. With this code if `connection` is `undefined`, it will still throw an error.

M-TGH: Truthy check will do as well, and it's probably easier to read, will refactor it. 👍 

However, since it was a loose equality check undefined wouldn't pass either as undefined and null are loosely equal. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness (JavaScript works in mysterious ways 😛)

chrisradek: Oh I see, I read that as a !== instead of !=.

PR #123, opened 2019-04-24
refactor: remove lodash from runtime dependencies

M-TGH:
*Issue #, if available:* #99/#84 (continuation/retry)

*Description of changes:*

Removes use of lodash functions making lodash no longer required on runtime.
To address the issue of prototype no longer being copied this adds a utility function that in base functionality does what omit previously did.
To address my concerns on performance versus _.omit in #99, it seems lodash themselves have removed omit in v5.0.0 due to performance reasons, see https://github.com/lodash/lodash/issues/2930#issuecomment-272298477. So accordingly, this should be a performance upgrade even.

I've also made a small fix to the tests already present for flush here: https://github.com/aws/aws-xray-sdk-node/blob/2908ef650c228b51b074063f3d1174afe3b96d42/packages/core/test/unit/segments/segment.test.js#L433
Noticed that this test wasn't properly functioning when I added my test to explicitly check whether a prototype property was still available on the segment, this because the args is an array of arguments sent to the function, however before this chance the argument array was used as `sentSegment`.


haotianw465:
Thank you for your contribution. We will take a look at this PR as soon as possible. Sorry about the late response.

Imran99:
Any idea when this will be merged?

alfaproject:
For me, the worst dependency is `winston` but this would help my bundle size a little bit too. Any plans to get this reviewed/merged?

willarmiros:
Hi @M-TGH, 
It looks like this branch is failing some Travis CI tests, could you investigate and resolve that?

M-TGH:
Hey @willarmiros. 

Good thing the tests caught that, not sure how/why I completely overlooked the isEmpty checks conditionally adding the subsegments. 😨 

Also, this isn't a replacement which goes to the full extent that the isEmpty check from lodash goes to, as they check for a lot of extra things that aren't quite necessary in this scenario (in my opinion). Although it might not hurt to also let it pass if it's either null or undefined. (let me know if this should be added)

packages/core/lib/segments/attributes/subsegment.js
```diff
@@ -124,15 +117,15 @@ Subsegment.prototype.addPrecursorId = function(id) {
  */
 
 Subsegment.prototype.addAnnotation = function(key, value) {
-  if (!(isBoolean(value) || isString(value) || isFinite(value))) {
+  if (!(typeof value === 'boolean' || typeof value === 'string' || (typeof value === 'number' && isFinite(value)))) {
```
willarmiros: Still has `isFinite` use

packages/core/lib/segments/segment_utils.js
```diff
@@ -1,5 +1,3 @@
-var isFinite = require('lodash/isFinite');
```
willarmiros: Another usage of `isFinite` on line 38.

packages/core/lib/segments/segment.js
```diff
@@ -77,13 +70,13 @@ Segment.prototype.addIncomingRequestData = function addIncomingRequestData(data)
  */
 
 Segment.prototype.addAnnotation = function addAnnotation(key, value) {
-  if (!(isBoolean(value) || isString(value) || isFinite(value))) {
+  if (typeof value !== 'boolean' && typeof value !== 'string' && !isFinite(value)) {
```
willarmiros: Another use of `isFinite`

packages/core/lib/segments/segment.js
```diff
@@ -346,18 +339,30 @@ Segment.prototype.close = function(err, remote) {
...
+    SegmentEmitter.send(thisCopy);
   }
 };
 
 Segment.prototype.format = function format() {
-  var ignore = ['counter', 'notTraced', 'exception'];
+  var ignore = ['segment', 'parent', 'counter'];
 
-  if (isEmpty(this.subsegments))
+  if(this.subsegments && this.subsegments.length === 0)
```
willarmiros: `isEmpty` from Lodash [returns true](https://lodash.com/docs/#isEmpty) if passed a `null` object or `undefined` value, so please mimic that behavior in all cases of replacing isEmpty.

packages/core/test/unit/segments/segment.test.js
```diff
@@ -430,11 +430,24 @@ describe('Segment', function() {
         segment.flush();
 
         sendStub.should.have.been.calledOnce;
-        var sentSegment = sendStub.lastCall.args;
+        var sentSegment = sendStub.lastCall.args[0];
         assert.notProperty(sentSegment, 'exception');
         assert.notProperty(sentSegment, 'counter');
         assert.notProperty(sentSegment, 'notTraced');
       });
+
```
willarmiros: Are you expanding upon one of our tests here? Not sure I understand this change in the context of this PR.

M-TGH: Yes this is a test added to ensure prototype properties on segments that are emitted are preserved. See PR body for more detail:

> I've also made a small fix to the tests already present for flush here:  https://github.com/aws/aws-xray-sdk-node/blob/2908ef650c228b51b074063f3d1174afe3b96d42/packages/core/test/unit/segments/segment.test.js#L433
> Noticed that this test wasn't properly functioning when I added my test to explicitly check whether a prototype property was still available on the segment, this because the args is an array of arguments sent to the function, however before this chance the argument array was used as `sentSegment`.

packages/core/lib/segments/attributes/subsegment.js
M-TGH: `isFinite` should be available natively, see the following MDN reference for more info: [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite)

As far as I'm aware these should be completely interchangeable.

packages/core/lib/segments/segment.js
M-TGH: Addressed in 5f9ab02.

packages/core/lib/segments/attributes/subsegment.js
willarmiros: I see, thanks for pointing that out!

packages/core/test/unit/segments/segment.test.js
willarmiros: I see. Apologies for glossing over that.

Issue #136, opened 2019-05-08
Unable to connect to mysql after upgrading to 2.3.3


rrosenblum:
After upgrading from version 2.2.0 to 2.3.3, I am unable to connect to mysql. I believe this is related to #98. There isn't much of an actual error being produced that I can see. The lambda just winds up timing out without ever actually connecting to mysql. I have tried increasing the timeout. I think this might be related to #98.

I have tried 2 different ways to capture mysql.

```js
const AWSXRay = require('aws-xray-sdk-core');    
const captureMySQL = require('aws-xray-sdk-mysql');    
const mysql = captureMySQL(require('mysql'));    
                                         
const pool = mysql.createPool({    
  host: process.env.DB_PRIMARY_ENDPOINT,    
  database: process.env.DB_NAME,    
  port: process.env.DB_PORT,    
  user: process.env.DB_USERNAME,    
  password: process.env.DB_PASSWORD    
});

module.exports.handler = async (event) => {
  pool.query
}
```
    
```js
const mysql = require('aws-xray-sdk').captureMySQL(require('mysql'));    
    
const pool = mysql.createPool({    
  host: process.env.DB_PRIMARY_ENDPOINT,    
  database: process.env.DB_NAME,    
  port: process.env.DB_PORT,    
  user: process.env.DB_USERNAME,    
  password: process.env.DB_PASSWORD    
});

module.exports.handler = async (event) => {
  pool.query
}
```

awssandra:
Hi rrosenblum,

Sorry for the issues. We'll do some testing on our end to see what the issue is. I think it's a fair call that the mysql2 changes may have broken the mysql patch functionality.

M-TGH:
Hey,

Thanks for the report. My apologies for any inconveniences. I didn't have too much time so far to look into it, but I was able to reproduce it. For now it should be alright to (temporarily) remove the parts that patch a Pool it's getConnection function. 

So far it looked like it's somewhere down that line that something was hitting a snag, maybe because the MySQL package internally uses getConnection on the Pool when using Pool.query. Unfortunately I couldn't yet figure out the full problem, but I'll try looking into it more as soon as I have the time. 

Suppose it might be interesting to see if this can be replicated with a test which stubs/simulates a Pool's query function.

Let me know if I can be of any more help.

M-TGH:
[...]

PR #62, opened 2018-10-23
simple support for mysql2 promise version of query

jafl:
*Issue #, if available:*

Missing support for mysql2 promises.

*Description of changes:*

The change in `patchCreateConnection()` detects a Promise and patches through to the actual Connection object.

The change in `resolveArguments()` adds support for passing `{sql,values}` as a single argument to `query()`, which is supported by mysql2.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #137, opened 2019-05-09
fix: keep Query instances passed to connection.query intact

M-TGH:
*Issue #, if available:* #136

*Description of changes:*

Preserves the first argument of resolveArguments when it is an Object and no callback(2nd argument) has been provided. This because in the case described in https://github.com/aws/aws-xray-sdk-node/issues/136#issuecomment-490763147 it's relying on the eventEmitter that's in the Mysql.Query instance provided in the first argument.

Since resolveArguments part where the first argument is an object part was added in #62, I would like to ask @jafl whether he remembers what case (presumably in mysql2) this was added, so we can check whether that would still pass as well.


jafl:
I think I simply missed that case.

PR #140, opened 2019-05-30
Reduce aws-sdk version down to the version packaged with Lambda 8.10

mdlavin:
When people deploy to AWS Lambda, they often exclude uploading the aws-sdk package since the Lambda runtime already includes a copy. I've been doing this for a while now and the AWS X-Ray package works fine. The only minor problem is that npm can complain when trying for force all packages to use the Lambda aws-sdk version using "resolutions" in package.json because the x-ray library claims to need a higher version.

I don't believe this package needs the higher version it claims to, and so reducing the package version will make it clear that this library is compatible with Lambda Node 8.10's builtin runtime.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -15,7 +15,7 @@
   },
   "dependencies": {
     "atomic-batcher": "^1.0.2",
-    "aws-sdk": "^2.304.0",
+    "aws-sdk": "^2.290.0",
     "continuation-local-storage": "^3.2.0",
     "date-fns": "^1.29.0",
     "lodash": "^4.17.11",


awssandra:
Looks good, we just have to verify the X-Ray sampling APIs are available on 2.290.0, vs the 2.304.0 version. Those aren't used by Lambda, but because they're a part of core it makes a difference.

haotianw465:
Hi, unfortunately we will not be able to accept this PR. The X-Ray SDK requires aws-sdk `2.304.0` at minimum for non-lambda feature: service managed sampling rules. You can see per https://github.com/aws/aws-sdk-js/blob/master/CHANGELOG.md for `2.304.0`: "feature: XRay: Support for new APIs that enable management of sampling rules."

The main issue here is that the  X-Ray SDK should have better modular design so the feature only for non-lambda applications is decoupled from lambda ones. This is on our backlog but we don't have an ETA. A workaround is to use Lambda layers so that the dependency of aws-sdk and aws-xray-sdk is handled separately.

Please let me know if you have any other concerns.

PR #144, opened 2019-06-22
Return response when using captureFunc and captureAsyncFunc

aldex32:
Hi guys,

I have created this PR in order to get the function result when capturing functions and async functions.
Let me know if this make sense to you.

Cheers,
Aldo

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -17,12 +17,13 @@ mechanisms, but a few are supplied.
 These keep the current subsegment up to date in automatic mode, or propagate the current subsegment in manual mode.
 
 `AWSXRay.captureFunc` - Takes a function that takes a single subsegment argument. This creates a new nested subsegment and exposes it. The segment
-closes automatically when the function finishes executing. This does not correctly
+closes automatically when the function finishes executing and returns the result if any. This does not correctly
 time functions with asynchronous calls. Instead, use
 captureAsyncFunc.
 
-`AWSXRay.captureAsyncFunc` - Takes a function that takes a single subsegment argument. This creates a new nested subsegment and exposes it. The segment
-must be closed using subsegment.close() or subsegment.close(error) when the asynchronous function completes.
+`AWSXRay.captureAsyncFunc` - Takes an async function that takes a single subsegment argument and returns the promise by executing the function. 
+This creates a new nested subsegment and exposes it. 
+The segment must be closed using subsegment.close() the asynchronous function completes successfully.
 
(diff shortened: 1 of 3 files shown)


awssandra:
Hi aldex32,

Thanks for the PR. Can you tell me a bit about the use case of this change? This would also require an edit to the API documentation/readme in respect to the functions behavior.

Thanks!
Sandra

aldex32:
Hi @awssandra ,

the use case would be when the user needs to trace a function which returns a value.
For instance I would like sub segment by wrapping an external library which does an http call and returns a result.
```
const getInfo = () => externalLibrary.getInfo();
const response = await AWSXRay.captureAsyncFunc('getInfo', getInfo, AWSXRay.getSegment())
```
If that makes sense for you then I can modify the API documentation.

Let me know,
Aldo

aldex32:
Hi @awssandra,

I didn't hear anymore from you. Do you agree with this PR?

williamhoyle:
@aldex32, your async example won't work as expected. You'd have to wrap the external function to close the segment:

```javascript
const getInfo = segment => externalLibrary
  .getInfo()
  .then(res => {
    segment.close();
    return res;
  })
  .catch(e => {
    segment.close(e);
    throw e;
  });
```

aldex32:
Hi @williamhoyle, 
I think you are right, but you don't need to close the segment in the catch, since it will be eventually closed by *captureAsyncFunc()*. So it should be something like this:
```JavaScript
const getInfo = async (segment) => {
  const result = await externalLibrary.getInfo();

  segment.close();

  return result;
}

const response = await AWSXRay.captureAsyncFunc('getInfo', getInfo, AWSXRay.getSegment());
```

williamhoyle:
@aldex32 eventually, but not with the correct timestamp. 

This is a good  PR without the async stuff. `captureFunc` can return a result without  side affects,  `captureAsyncFunc` not so much. 

Just back the async stuff out and I think you've got a shot.

haotianw465:
Hi @williamhoyle , what do you mean by there are side effects on `captureAsyncFun`? It looks to me the timestamp is accurate on both success and exception.

williamhoyle:
> Hi @williamhoyle , what do you mean by there are side effects on `captureAsyncFun`? It looks to me the timestamp is accurate on both success and exception.

Sorry, my comment was poorly worded. I was referring to @aldex32's example, not the current implementation of `captureAsyncFun`.

haotianw465:
Thank you for your contribution. I have merged the change to master.

aldex32:
Thanks for merging it!

PR #146, opened 2019-06-26
Fixing eslint errors in packages. Updated .eslintrc.json to enable es6.

srprash:
*Issue #, if available:*

*Description of changes:*
This PR fixes the following errors which were reported by eslint:
- indentations
- missing semicolons
- undefined variables
- invalid use of named params
- removing unused variables
- " to '

### DIFF
--- a/packages/core/.eslintrc.json
+++ b/packages/core/.eslintrc.json
@@ -1,7 +1,8 @@
 {
     "env": {
         "node": true,
-        "mocha": true
+        "mocha": true,
+        "es6": true
     },
     "extends": "eslint:recommended",
     "rules": {
--- a/packages/core/lib/logger.js
+++ b/packages/core/lib/logger.js
@@ -1,5 +1,5 @@
 var winston = require('winston');
-var format = require('date-fns/format')
+var format = require('date-fns/format');
 
 var logger;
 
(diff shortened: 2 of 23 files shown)


packages/core/.eslintrc.json
```diff
@@ -1,7 +1,8 @@
 {
     "env": {
         "node": true,
-        "mocha": true
+        "mocha": true,
+        "es6": true
```
luluzhao: Just wondering what is the benefit about enable es6?

packages/core/lib/logger.js
```diff
@@ -1,5 +1,5 @@
 var winston = require('winston');
-var format = require('date-fns/format')
+var format = require('date-fns/format');
```
luluzhao: Even if there is no ";" presenting at the end of each line, the code will still compile without throwing an error before?

srprash: Yes. Semicolon is optional in Javascript and the code does compile and runs with or without it. However, as a best practice, it is advised to put use semicolons.

packages/core/.eslintrc.json
srprash: Setting `es6` to `true` enables all of the ES6 globals such as Promise, Map, Set, etc. If not enabled, the eslint linter will raise the 'Promise not defined' error.

PR #172, opened 2019-08-28
Add log level setting

### LINKED ISSUE #163: Logging issue
I've followed the discussion in https://github.com/aws/aws-xray-sdk-node/issues/2 and tried to set my own logger at a level of error to stop receiving the following logs without any success.

`
AWS_XRAY_DAEMON_ADDRESS is set. Configured daemon address to 169.254.79.2:2000.
AWS_XRAY_CONTEXT_MISSING is set. Configured context missing strategy to LOG_ERROR.
Subsegment streaming threshold set to: 0
`

We setup the X-Ray as below
`
const AWSSDK = require('aws-sdk');
const AWSXRay = require('aws-xray-sdk-core');
AWSXRay.setLogger(loggerForXRAY);
AWSXRay.captureAWS(AWSSDK)
`

Is this something I've approached in the wrong way or is there some other issue?

grahamhar:
Attempt to fix #163 
In the lambda execution environment having xray info logs adds noise and
potentially cost, this add the ability to set the log level via
environment variable allowing these to be suppressed.

### DIFF
--- a/packages/core/test/unit/logger.test.js
+++ b/packages/core/test/unit/logger.test.js
@@ -1,0 +1,39 @@
+var assert = require('chai').assert;
+
+var logger = require('../../lib/logger');
+
+describe('logger', function () {
+  function reloadLogger() {
+    var path = '../../lib/logger';
+    delete require.cache[require.resolve(path)];
+    logger = require(path);
+  }
+
+  describe('defaults', function () {
+    afterEach(function () {
+      delete process.env.AWS_XRAY_DEBUG_MODE;
+      reloadLogger();
+
+    });
+    it('Should have a default logger with no transport and level set to info', function () {
+      assert.deepEqual(logger.getLogger().transports, {});
+      assert.equal(logger.getLogger().level, 'info');
+    });
+    it('Should have a logger with console transport with level set to debug when AWS_XRAY_DEBUG_MODE is set', function () {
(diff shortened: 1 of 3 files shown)


packages/core/test/unit/logger.test.js
```diff
@@ -0,0 +1,39 @@
...
+    logger = require(path);
+  }
+
+  describe('defaults', function () {
+    afterEach(function () {
+      delete process.env.AWS_XRAY_DEBUG_MODE;
+      reloadLogger();
+
+    });
+    it('Should have a default logger with no transport and level set to info', function () {
```
willarmiros: This test fails if the machine from which it is run has the 'AWS_XRAY_LOG_LEVEL' environment variable set already. I recommend setting it equal to the empty string then reloading the logger as you do in other tests before the assertions.

grahamhar: Good spot, I've addressed this I think.

Also I think the same scenario could have happened in the final test so explicitly unset AWS_XRAY_DEBUG_MODE there.

willarmiros: Tests pass in all cases now. Thanks for the fix.

grahamhar:
@willarmiros do you have any idea when this will be available in published version?

willarmiros:
Unfortunately I cannot provide a timeline on future releases.

PR #187, opened 2019-09-16
made aws-sdk a dev-dependency

joshuawinsor:
*Issue #, if available:*
186

*Description of changes:*
made aws-sdk a devDependency

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -15,7 +15,6 @@
   },
   "dependencies": {
     "atomic-batcher": "^1.0.2",
-    "aws-sdk": "^2.304.0",
     "continuation-local-storage": "^3.2.0",
     "date-fns": "^1.29.0",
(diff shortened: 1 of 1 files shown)


willarmiros:
Hello,

We recognize that the `aws-sdk` dependency greatly inflates the size of the package, however it is used for centralized sampling [here](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/middleware/sampling/service_connector.js#L2). This is an integral part of our customer experience for customers using EC2. While this may be frustrating for Lambda customers because the AWS SDK is available as a layer, we cannot remove support for centralized sampling just to reduce this package size. I'm closing this PR for now.

On our backlog is an item to refactor centralized sampling into a separate module that would allow only customers (like those on EC2) to include it while others who don't use it (like those on Lambda) can exclude the `aws-sdk` dependency.

Issue #103, opened 2019-02-28
Critical dependency warning when bundling XRay SDK with webpack


jhecking:
Since adding aws-xray-sdk-core to my package dependencies, I'm getting the following warnings when bundling my package using webpack:

```
WARNING in ./node_modules/aws-xray-sdk-core/lib/patchers/call_capturer.js 41:32-47
Critical dependency: the request of a dependency is an expression
 @ ./node_modules/aws-xray-sdk-core/lib/segments/attributes/aws.js
 @ ./node_modules/aws-xray-sdk-core/lib/aws-xray.js
 @ ./node_modules/aws-xray-sdk-core/lib/index.js
 @ ./node_modules/aws-xray-sdk/lib/index.js
 @ ./src/util/xray.ts
 @ ./src/abc/def/xxx.ts
 @ multi ./source-map-install.js ./src/abc/def/xxx.ts

WARNING in ./node_modules/colors/lib/colors.js 127:29-43
Critical dependency: the request of a dependency is an expression
 @ ./node_modules/colors/safe.js
 @ ./node_modules/winston/lib/winston/config.js
 @ ./node_modules/winston/lib/winston.js
 @ ./node_modules/aws-xray-sdk-core/lib/logger.js
 @ ./node_modules/aws-xray-sdk-core/lib/context_utils.js
 @ ./node_modules/aws-xray-sdk-core/lib/aws-xray.js
 @ ./node_modules/aws-xray-sdk-core/lib/index.js
 @ ./node_modules/aws-xray-sdk/lib/index.js
 @ ./src/util/xray.ts
 @ ./src/abc/def/xxx.ts
 @ multi ./source-map-install.js ./src/abc/def/xxx.ts
```

Webpack tries to resolve require calls statically to make a minimal bundle. When a library uses variables in a require call (such as [this line](https://github.com/aws/aws-xray-sdk-node/blob/d37d2d5dda348f263ca555389a6613ee2ae834be/packages/core/lib/patchers/call_capturer.js#L41) in aws-xrad-sdk-core), Webpack cannot resolve them statically and imports the entire package.

I'm using v2.2.0 of the SDK.

chrisradek:
@jhecking 
Thank you for reporting this. I'll have to dig into why we used a `require` to load a JSON file instead of the `fs` module, since changing that now could break if users were passing in a node module that contained their JSON.

[...]

Issue #48, opened 2018-07-24
Incompatible with Winston v3.0


IllegalCactus:
Using the SDK with Winston v.3.0.0 produces the following error:

```
node_modules/aws-xray-sdk-core/lib/logger.js:17
    logger = new (winston.Logger)({});

TypeError: winston.Logger is not a constructor
```

Winston had a breaking change in 3.0 which changes how the logger should be constructed:

https://github.com/winstonjs/winston#creating-your-own-logger

https://github.com/winstonjs/winston/commit/eff6f00588ea6e7fcfc0a33bcf049ad85c76b6d9#diff-fc1a687308b45c41cf7efa7aff4253dd

Not sure if updating the Winston version in this project is an option?

awssandra:
Hi Cactus,

We're planning on removing Winston - see https://github.com/aws/aws-xray-sdk-node/issues/10. However, in the mean time, if you'd like to submit a PR with a bump to 3.0 we'd be happy to review it.

Is there any particular reason you're wanting to use the SDK with Winston 3?

IllegalCactus:
Ah I didn't see that issue, thanks. In that case, I'm happy to close this as I don't have an urgent need to use 3.0 in the meantime. I just spotted this error while upgrading some dependencies in a project but in my case I can stick with 2.x too.

Issue #10, opened 2017-11-27
Optimise Dependencies


adieuadieu:
Hi there. Thank you for your work on this project. X-Ray is great!

I have some concerns about some of the dependency decisions made in the project. I'd like to use the `aws-xray-sdk-core` module in a Lambda handler, however simply including this library adds around 900 KB of code. That's crazy!

There must be something we can do to reduce the code size. For example, does the package really need to [depend on](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/logger.js#L29) Moment.js? Also, is it necessary to [inline](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/logger.js#L43) polyfills? Is `winston` really necessary?

Let me know your thoughts. I'm willing to raise a PR with some improvements (like removing Moment, for example.)

awssandra:
Hi adieuadieu,

Winston and Moment are used for formatting and configuring all the SDK output.  They are only used to a) capture and format timing information and b) use log levels appropriately. Ideally, we would replace these with our own baked in minimal solution, and create our own base logger object that continues to support log levels and timestamps without the bloated dependencies. If you'd like to open a PR we would be happy to review your changes.

Thanks,
Sandra

jplock:
@awssandra based on https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/logger.js#L21, would you welcome a PR to just remove `winston` entirely?  I'm not sure what value it is providing considering the size it adds to the Lambda's when included.

jplock:
I'll see if I can submit a PR to remove winston for review.

haotianw465:
Thank you for your deep dive and contribution @vladgolubev. I have merged your PR.

We will keep the action item of upgrading Winston3 or a better replacement if possible.

jplock:
@haotianw465 is it possible to create a new `aws-xray-sdk-node` release after the winston change?

vladgolubev:
@jplock just in case there will be no new release, this is can be fixed if you are using yarn via `resolutions` in `package.json`

```
"resolutions": {
	"winston": "2.4.4"
}
```

See https://yarnpkg.com/lang/en/docs/selective-version-resolutions/

jplock:
Thanks @vladgolubev but we're using npm.  I'll wait until the next published released as I don't want to reference GitHub.

PR #188, opened 2019-09-17
Upgrade of Winston library from version 2.4.4 to 3.2.1

briancullen:
*Issue #, if available:*
#10 #48 #103 

*Description of changes:*
Upgrade of Winston library from version 2.4.4 to 3.2.1

I tried to keep the changes to a minimum and hence configured the new version to get the behaviour as close as possible to that of the previous version. As part of this change the date-fns package was removed.

There are additional changes/issues that might be worth considering. First, I can't see anything adding meta information to the logs so it might be better to remove the metadata formatter.

Second, it looks like the logging for lambda is treated differently (see lines 31 to 38 of logger.js) and I'm not sure why. In particular it would seem that this code removes any filtering on log level.

Third, as mentioned in one of the issues, using winston might be overkill for what we want to achieve. Maybe it would just be better to have a standard logger that writes everything to the console and then let users to override if they want? 

### DIFF
--- a/packages/core/lib/logger.js
+++ b/packages/core/lib/logger.js
@@ -1,5 +1,4 @@
 var winston = require('winston');
-var format = require('date-fns/format');
 
 var logger;
 var xrayLogLevel = process.env.AWS_XRAY_LOG_LEVEL;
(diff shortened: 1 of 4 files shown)


willarmiros:
Hi Brian,
Thank you very much for your contribution. We will be taking a look shortly.

briancullen:
In the initial comment I mentioned that it might be preferable to remove Winston as a dependency as the module doesn't use much of it's functionality. Pull request #190 was created to show this different approach.

briancullen:
I will leave this open until certain but it appears that this will be dropped in favour of #193

PR #191, opened 2019-09-23
Get protocol using standard ClientRequest interface

gugu:
*Description of changes:*

According to node.js documentation nobody promises that ClientRequest will have "agent" attribute. For example, patching @hapi/wreck fails, because ClientRequest does not have attribute "agent". It is safer to use ClientRequest.connection.secure, which is a part of ClientRequest standard API


willarmiros:
Hi @gugu,

Thank you for your contribution. I would like to request that you make this PR against the `2.x` [branch](https://github.com/aws/aws-xray-sdk-node/tree/2.x) instead of master. See #192 for more details. It also looks like your change is not building due to a missing parenthesis. 

Sorry for this inconvenience, you can close this PR and we'll follow up on the new one once it's created and the bug is fixed. Thank you again.

PR #109, opened 2019-04-03
Add TypeScript type definitions (in progress)

Kroisse:
*Issue #, if available:* #14 

*Description of changes:*

Adds type definitions of "public" items of this library, for TypeScript compiler.

This was derived from my private work, so lots of items are omitted or incomplete.

*Todo:*

 - [ ] Add type inference tests using [dtslint](https://github.com/Microsoft/dtslint)
   - See also: https://hackernoon.com/testing-types-an-introduction-to-dtslint-b178f9b18ac8
 - [ ] Add `XRaySegment` field in the parameter of all methods of capturing clients, for manual mode.
 - [ ] Add complete set of type definitions of `aws-xray-sdk-core`
 - [ ] Add a type definition of `aws-xray-sdk-express`
 - [ ] Add a type definition of `aws-xray-sdk-mysql`
 - [ ] Add a type definition of `aws-xray-sdk-postgres`
 - [ ] Add a type definition of `aws-xray-sdk-restify`
 - [ ] Add a type definition of `aws-xray-sdk`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #14, opened 2018-01-06
Typescript Definitions


glicht:
Hi,

Are there anywhere typescript definitions for this project? 
If not, do you have any plans to provide typescript definitions in the future?

thanks,

Guy

haotianw465:
Hi, the SDK currently doesn't have typescript definitions available yet. We will looking into providing the definitions but we cannot guarantee such offering nor a timeline. Please stay tuned.

piersmacdonald:
Lacking Typescript and Node 8+ support isn't great for a non-beta AWS product.

NachtmahrEDM:
I also would love to see typescript definitions soon as our company is dependent on aws and typescript.

voxpelli:
Definitions are getting increasingly common among projects and since Typescript at least nowadays can also be used to lint pure js I think it's a very good suggestion.

However: Anyone can contribute type definitions for this package to the @DefinitelyTyped project: https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package

alex-at-cascade:
This would be very nice to have, as it's very frustrating how the `AWSXRay.captureAWS(...)` call obscures the valuable `.d.ts` declarations provided by the AWS sdk.

kerbyferris:
As a workaround, generics can help. Creating a `aws-xray-sdk.d.ts` types file with something like the following will satisfy the compiler without obscuring AWS sdk declarations:

```
declare module "aws-xray-sdk" {
  export function captureAWSClient<T>(client: T): T;
  ...
}
```

sergiovm:
> Definitions are getting increasingly common among projects and since Typescript at least nowadays can also be used to lint pure js I think it's a very good suggestion.
> 
> However: Anyone can contribute type definitions for this package to the @DefinitelyTyped project: https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package

Somebody created a basic definition here:
https://github.com/mziyabo-aws/aws-xray-typings

Not perfect though...

woodsae:
I would really like to see this done as well.

esilkensen:
Hi everyone, I've just opened #207  🙂

willarmiros:
TypeScript definitions have been released in v2.5.0!

LosD:
@willarmiros Is it on purpose that the definitions has been added to dependencies instead of dev-dependencies? That seems like a mistake.

esilkensen:
[...]

Issue #157, opened 2019-07-23
Node SDK release V3.0.0


yogiraj07:
Hello,
This issue tracks next major version release changes : `V3.0.0` 

## 1) Using `cls-hooked` library for trace context
Currently, `master` branch uses `continuation-local-storage` library for thread local storage. `cls-hooked` library is used as thread local storage for async operations in the `expermental-hooks` [branch](https://github.com/aws/aws-xray-sdk-node/blob/experimental-hooks/packages/core/lib/context_utils.js#L5). The next major version will merge branch `expermental-hooks`  into `master` and `cls-hooked` library will be used for synchronous and asynchronous operations. `expermental-hooks` branch will be deprecated.

When running Nodejs version >= 8.2.1, `cls-hooked` module uses the newer async_hooks API which is considered Experimental by Nodejs : [reference](https://www.npmjs.com/package/cls-hooked#when-running-nodejs-version--821-this-module-uses-the-newer-async_hooks-api-which-is-considered-experimental-by-nodejs). 

This change is included as per community [feedback](https://github.com/aws/aws-xray-sdk-node/issues/60#issuecomment-496333896).

### Action required for update
It is unlikely this change will  cause any existing code to break. If you'd like to be safe, test your application with the `experimental-hooks` branch of the SDK. Installation instructions can be found [here](https://github.com/aws/aws-xray-sdk-node/tree/experimental-hooks). The only difference between the `2.x` and `experimental-hooks` branches is the use of `cls-hooked`, so this would be the best way to detect if the change causes any issues in your application.

## 2) Dropping support for Node V4, V6, V8
Currently the SDK supports Node versions 4, 6, and 8. Since these versions have met end of life : [reference](https://nodejs.org/en/about/releases/), the major version will drop support for these versions.

### Action required for update
Ensure your application uses Node 10 or above.

## 3) Removal of winston
We will merge #190 to remove the dependency on the `winston` logging library. This will reduce the size of our bundle, a big customer ask (see #113). `winston` is an overly complex and cumbersome library given our simple use case, and it will be replaced with a lightweight, native console logger. All logs emitted from the SDK will be formatted the same way as

     [timestamp] [log level] [message] [optional metadata]

The `setLogger()` API will be unchanged, so you can still replace the default logger with any logger object of your choice as long as it implements the debug, info, warn, and error methods.

### Action required for update
[...]

PR #207, opened 2019-10-30
Add TypeScript definitions

esilkensen:
*Issue #, if available:* #14 

*Description of changes:*

This PR makes a complete first pass at adding TypeScript definitions for aws-xray-sdk. It leverages some of the work done in a previous PR that seems to be abandoned (#109, thank you to that author!)

I realize it's a pretty big PR, so I will be happy to do anything I can that might make it easier to review, and of course to iterate on it in response to any feedback.

It adds .d.ts files next to the corresponding .js files, and a `"types": "lib/index.d.ts"` entry to each package.json.

For testing, it uses the [`tsd`](https://github.com/SamVerschueren/tsd) tool which by default looks for TypeScript files to check in a `test-d` directory; each package has a `test-d/index.test-d.ts` file which exercises its TypeScript definitions, and the `test` script has been updated to run `tsd` as well.

Some initial questions I have:
* Is it exporting types that aren't necessarily meant to be part of the public API?
* How best to manage dependencies on `@types` packages for express/mysql/postgres?

---

### DIFF
--- a/packages/core/lib/middleware/incoming_request_data.d.ts
+++ b/packages/core/lib/middleware/incoming_request_data.d.ts
@@ -1,0 +1,11 @@
+import * as http from 'http';
+
+declare class IncomingRequestData {
+  request: { [key: string]: any };
+
+  constructor(req: http.IncomingMessage);
+
+  close(res: http.ServerResponse): void;
+}
+
+export = IncomingRequestData;
--- a/packages/core/lib/middleware/mw_utils.d.ts
+++ b/packages/core/lib/middleware/mw_utils.d.ts
@@ -1,0 +1,60 @@
+import * as http from 'http';
+import Segment = require('../segments/segment');
+import IncomingRequestData = require('./incoming_request_data');
+
+export const defaultName: string | undefined;
+
+export const dynamicNaming: boolean;
+
+export const hostPattern: string | null;
+
+export function enableDynamicNaming(hostPattern?: string): void;
+
+export function processHeaders(req?: Partial<Pick<http.IncomingMessage, 'headers'>>): { [key: string]: string };
+
+export function resolveName(hostHeader?: string): string;
+
+export function resolveSampling(
+  amznTraceHeader: { [key: string]: string },
+  segment: Segment,
+  res: http.ServerResponse
+): void;
+
+export function setDefaultName(name: string): void;
+
+export function disableCentralizedSampling(): void;
+
+export interface BaseRuleConfig {
+  http_method: string;
+  url_path: string;
+  fixed_target: number;
+  rate: number;
+  description?: string;
+}
+
+export interface RuleConfigV1 extends BaseRuleConfig {
+  service_name: string;
+}
+
+export interface RuleConfigV2 extends BaseRuleConfig {
+  host: string;
+}
+
+export type RuleConfig = RuleConfigV1 | RuleConfigV2;
+
+export interface DefaultRuleConfig {
+  fixed_target: number;
+  rate: number;
+}
+
+export interface RulesConfig {
+  version: number;
+  default: DefaultRuleConfig;
+  rules?: RuleConfig[];
+}
+
+export function setSamplingRules(source: string | RulesConfig): void;
+
+export {
+  IncomingRequestData
+}
--- a/packages/core/lib/patchers/aws_p.d.ts
+++ b/packages/core/lib/patchers/aws_p.d.ts
@@ -1,0 +1,38 @@
+import * as AWS from 'aws-sdk';
+import Segment = require('../segments/segment');
+import Subsegment = require('../segments/attributes/subsegment');
+
+export type Callback<D> = (err: AWS.AWSError | undefined, data: D) => void;
+
+export interface AWSRequestMethod<P, D> {
+  (params: P, callback?: Callback<D>): AWS.Request<D, AWS.AWSError>;
+  (callback?: Callback<D>): AWS.Request<D, AWS.AWSError>;
+}
+
+export type PatchedAWSRequestMethod<P, D> = AWSRequestMethod<P & { XRaySegment?: Segment | Subsegment }, D>;
+
+export type PatchedAWSClient<T extends AWS.Service> = {
+  [K in keyof T]: T[K] extends AWSRequestMethod<infer P, infer D>
+  ? PatchedAWSRequestMethod<P, D>
+  : T[K]
+};
+
+export interface AWSClientConstructor<P, T extends typeof AWS.Service> {
+  new(params?: P): InstanceType<T>;
+}
+
+export interface PatchedAWSClientConstructor<P, T extends typeof AWS.Service> {
+  new(params?: P): PatchedAWSClient<InstanceType<T>>;
+}
+
+export type PatchedAWS<T = typeof AWS> = {
+  [K in keyof T]: T[K] extends typeof AWS.Service
+  ? (T[K] extends AWSClientConstructor<infer P, T[K]>
+    ? PatchedAWSClientConstructor<P, T[K]>
+    : T[K])
+  : T[K];
+};
+
+export function captureAWS(awssdk: typeof AWS): PatchedAWS;
+
+export function captureAWSClient<T extends AWS.Service>(service: T): PatchedAWSClient<T>;
--- a/packages/core/lib/segments/attributes/subsegment.d.ts
+++ b/packages/core/lib/segments/attributes/subsegment.d.ts
@@ -1,0 +1,56 @@
+import * as http from 'http';
+
+declare class Subsegment {
+  id: string;
+  name: string;
+  start_time: number;
+  in_progress?: boolean;
+
+  constructor(name: string);
+
+  addNewSubsegment(name: string): Subsegment;
+
+  addSubsegment(subsegment: Subsegment): void;
+
+  removeSubsegment(subsegment: Subsegment): void;
+
+  addAttribute(name: string, data: any): void;
+
+  addPrecursorId(id: string): void;
+
+  addAnnotation(key: string, value: boolean | string | number): void;
+
+  addMetadata(key: string, value: any, namespace?: string): void;
+
+  addSqlData(sqlData: any): void;
+
+  addError(err: Error | string, remote?: boolean): void;
+
+  addRemoteRequestData(req: http.ClientRequest, res: http.IncomingMessage, downstreamXRayEnabled: boolean): void;
+
+  addFaultFlag(): void;
+
+  addErrorFlag(): void;
+
+  addThrottleFlag(): void;
+
+  close(err?: Error | string | null, remote?: boolean): void;
+
+  incrementCounter(additional?: number): void;
+
+  decrementCounter(): void;
+
+  isClosed(): boolean;
+
+  flush(): void;
+
+  streamSubsegments(): true | undefined;
+
+  format(): string;
+
+  toString(): string;
+
+  toJSON(): { [key: string]: any };
+}
+
+export = Subsegment;
--- a/packages/core/lib/segments/segment.d.ts
+++ b/packages/core/lib/segments/segment.d.ts
@@ -1,0 +1,60 @@
+import Subsegment = require('./attributes/subsegment');
+import IncomingRequestData = require('../middleware/incoming_request_data');
+
+declare class Segment {
+  id: string;
+  name: string;
+  start_time: number;
+  in_progress?: boolean;
+  trace_id: string;
+  parent_id?: string;
+  origin?: string;
+
+  constructor(name: string, rootId?: string | null, parentId?: string | null);
+
+  addIncomingRequestData(data: IncomingRequestData): void;
+
+  addAnnotation(key: string, value: boolean | string | number): void;
+
+  setUser(user: string): void;
+
+  addMetadata(key: string, value: any, namespace?: string): void;
+
+  setSDKData(data: object): void;
+
+  setMatchedSamplingRule(ruleName: string): void;
+
+  setServiceData(data: any): void;
+
+  addPluginData(data: object): void;
+
+  addNewSubsegment(name: string): Subsegment;
+
+  addSubsegment(subsegment: Subsegment): void;
+
+  removeSubsegment(subsegment: Subsegment): void;
+
+  addError(err: Error | string, remote?: boolean): void;
+
+  addFaultFlag(): void;
+
+  addErrorFlag(): void;
+
+  addThrottleFlag(): void;
+
+  isClosed(): boolean;
+
+  incrementCounter(additional?: number): void;
+
+  decrementCounter(): void;
+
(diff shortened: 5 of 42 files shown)


mmmeff:
Just curious - what's the reasoning for not adding these types to the `@types/*` monorepo? That's more of the standard approach in the typescript community as it allows the entire community to maintain type definitions without imposing maintenance of those types on a project's maintainers if they prefer plain js.

By checking these type definitions into this project, `aws-xray-sdk` now needs to maintain them as well, either directly or by repeatedly taking these kinds of PRs.

packages/core/lib/middleware/mw_utils.d.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export const defaultName: string | undefined;
+
+export const dynamicNaming: boolean;
+
+export const hostPattern: string | null;
+
+export function enableDynamicNaming(hostPattern?: string): void;
+
+export function processHeaders(req?: Partial<Pick<http.IncomingMessage, 'headers'>>): { [key: string]: string };
```
ArturGrigio: why `http.IncomingMessage` and not `http.IncomingMessage | https.IncomingMessage`?

packages/core/lib/middleware/incoming_request_data.d.ts
```diff
@@ -0,0 +1,11 @@
+import * as http from 'http';
+
+declare class IncomingRequestData {
+  request: { [key: string]: any };
+
+  constructor(req: http.IncomingMessage);
```
ArturGrigio: why is the `req` only typed as `http.IncomingMessage` and not `http.IncomingMessage | https.IncomingMessage`

packages/core/lib/middleware/mw_utils.d.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export function resolveName(hostHeader?: string): string;
+
+export function resolveSampling(amznTraceHeader: object, segment: Segment, res: http.ServerResponse): void;
```
ArturGrigio: Shouldn't the return type of `resolveSampling()` be a `boolean`?

esilkensen:
Hi @mmmeff, 

While DefinitelyTyped is a great resource, I think if aws-xray-sdk contributors are willing there are some advantages to bundling the type definitions with the package.

Of course, for the types to be useful they will have to be maintained regardless, and having them here would mean that could be done in one place, rather than opening corresponding PR's in DefinitelyTyped for every change to the public API.

I can see how it's an imposition, but note that it's not _technically_ one, at least as far as: the type definitions could be allowed to drift from the JS here the same as in DefinitelyTyped; there isn't anything that will break the build if the .js doesn't match the .ts.d.

For what it's worth, aws-sdk had a [@types/aws-sdk](https://www.npmjs.com/package/@types/aws-sdk) until version 2.7.0, and since then has been including types with aws-sdk itself -- though it is able to automatically generate some of them.

Anyway, if the consensus is that these definitions belong in DefinitelyTyped, I will open a PR there instead.

packages/core/lib/middleware/mw_utils.d.ts
esilkensen: 1. That sounds right, I'll update.
2. I don't think `https.IncomingMessage` exists? It's not in `@types/node`, and I don't see it in the source or docs https://nodejs.org/api/https.html
3. The JSDoc does say `@returns {boolean}`, but I think `void` matches the implementation: there is no `return` statement.

packages/core/lib/utils.d.ts
```diff
@@ -0,0 +1,19 @@
+import Segment = require('./segments/segment');
+
+export function getCauseTypeFromHttpStatus(status: number): 'error' | 'fault' | undefined;
```
ArturGrigio: `status` should be a `string`, not a `number`

packages/core/lib/segments/segment_utils.d.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function getCurrentTime(): number;
+
+export function setOrigin(origin: any): void;
+
+export function setPluginData(pluginData: any): void;
+
+export function setSDKData(sdkData: any): void;
+
+export function setServiceData(serviceData: any): void;
```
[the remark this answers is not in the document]
esilkensen: What about `object` for `serviceData`? It looks like this value gets passed to it in aws-xray.js https://github.com/aws/aws-xray-sdk-node/blob/180e70bbab811d0beb994b2804e4b832cbab8698/packages/core/lib/aws-xray.js#L345-L350

packages/core/lib/utils.d.ts
esilkensen: I was basing `number` off https://nodejs.org/api/http.html#http_response_statuscode and  https://github.com/aws/aws-xray-sdk-node/blob/180e70bbab811d0beb994b2804e4b832cbab8698/packages/core/test/unit/utils.test.js#L12-L20

But does `getCauseTypeFromHttpStatus` definitely get called with a `string` too? It could be `number | string`?

packages/core/lib/segments/segment.d.ts
```diff
@@ -0,0 +1,59 @@
+import Subsegment = require('./attributes/subsegment');
+import IncomingRequestData = require('../middleware/incoming_request_data');
+
+declare class Segment {
+  id: string;
+  name: string;
+  start_time: number;
+  in_progress?: boolean;
+  counter: number;
```
esilkensen: this is an example of where I wonder, should `counter` be public?

packages/core/lib/segments/segment.d.ts
```diff
@@ -0,0 +1,59 @@
...
+  trace_id: string;
+  parent_id?: string;
+  origin?: string;
+
+  constructor(name: string, rootId?: string | null, parentId?: string | null);
+
+  addIncomingRequestData(data: IncomingRequestData): void;
```
esilkensen: the implementation sets `this.http = data`, should this class have a public `http: IncomingRequestData`?

packages/core/lib/segments/attributes/subsegment.d.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  addAnnotation(key: string, value: boolean | string | number): void;
+
+  addMetadata(key: string, value: any, namespace?: string): void;
+
+  addSqlData(sqlData: any): void;
+
+  addError(err: Error | string, remote?: boolean): void;
+
+  addRemoteRequestData(req: http.ClientRequest, res: http.IncomingMessage, downstreamXRayEnabled: boolean): void;
```
esilkensen: the implementation sets `this.http = new RemoteRequestData(...)`, should this class have a public `http: RemoteRequestData`? I hadn't created a remote_request_data.d.ts yet.

packages/core/lib/utils.d.ts
ArturGrigio: the function signature expects a string, but seeing how the status code is a number, I'd agree that a `number | string` is the safest bet.

packages/core/lib/segments/segment_utils.d.ts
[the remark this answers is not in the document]
ArturGrigio: @willarmiros and I couldn't find a non-string call to `setServiceData()`, but I stand corrected.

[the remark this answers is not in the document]
esilkensen: think it should be... `object`, `string | object`, fall back to `any`?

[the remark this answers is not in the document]
ArturGrigio: I stand corrected. I think leaving it loose (with `any`) is a safe bet.

packages/core/lib/segments/segment.d.ts
ArturGrigio: You are correct, it could be a private attribute.

packages/core/lib/segments/segment.d.ts
willarmiros: It would be correct to have the optional `http` field, but I am unaware of any use cases that require the http element to be accessed publicly, so I'd recommend making it private.

esilkensen: Sounds good - what do you think about including private fields/methods/etc. marked as `private` vs. omitting them altogether as implementation details?

willarmiros: I was thinking about that - since there are no plans to make this SDK native TS anytime soon including private fields seems like somewhat wasted work. There's certainly no harm in adding them as private fields but I will approve the PR without them.

packages/core/lib/segments/attributes/subsegment.d.ts
willarmiros: Same as above - no need to include it.

packages/core/lib/segments/segment_utils.d.ts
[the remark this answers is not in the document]
willarmiros: @esilkensen that strategy sounds best.

willarmiros:
To answer your questions:
* The exported types seem to be fine as is, they line up with the exported functions and fields in the original package as far as I can tell. Were there any types in particular you were referring to? 
* Having the `@Types` for express etc as dependencies is fine. We're exporting types that are defined in those external packages, so it'll be necessary to have them as dependencies for consumers of `aws-xray-sdk`. 

Moving forward, we will be supporting TypeScript definitions in future releases of the SDK. The only foreseeable issue is that when we separate the AWS SDK dependency from the core module (item 4 in #157) we'll have to remove all TS definitions imported from `aws-sdk`. They'll probably just be replaced with `any`, which will degrade the usefulness of TS definitions in those places unless you see any ways around this.

As a final note, I am new to TypeScript (as you could tell by my reliance on @ArturGrigio to review  this). I'd appreciate if the community point out any misses on my end as I make changes to TS definitions during future maintenance.

packages/core/lib/patchers/aws_p.d.ts
```diff
@@ -0,0 +1,38 @@
...
+import Subsegment = require('../segments/attributes/subsegment');
+
+export type Callback<D> = (err: AWS.AWSError | undefined, data: D) => void;
+
+export interface AWSRequestMethod<P, D> {
+  (params: P, callback?: Callback<D>): AWS.Request<D, AWS.AWSError>;
+  (callback?: Callback<D>): AWS.Request<D, AWS.AWSError>;
+}
+
+export type PatchedAWSRequestMethod<P, D> = AWSRequestMethod<P & { XRaySegment?: Segment | Subsegment }, D>;
```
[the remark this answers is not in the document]
willarmiros: Copying the definitions over is a possibility, I will get in touch with the AWS SDK team to see what they recommend.

kwongkz:
May i know when this will release to npm?

willarmiros:
It will be included in the 2.5.0 release of the SDK. I cannot give an exact date but I will update this when it is released.

willarmiros:
v2.5.0 has been released to NPM.

esilkensen:
Awesome, thanks @willarmiros!

I was just looking again, should the `segment` argument to `AWSXRay.resolveSegment` be optional (i.e. for automatic mode)? Sorry, I think it may be something I missed in my PR, and the type lists it as required.

willarmiros:
I guess that was missed on our end as well. Hmm this is where my TS knowledge is lacking. It is always invoked with an argument in the segment slot, but when operating in automatic mode, `null` is passed in instead of a segment object. It looks like whether TS will complain depends on whether the `--strictNullChecks` option is used?

Regardless, it should definitely allow for `null` values, and also be optional in case any customers call it without args.

esilkensen:
As it is now, my understanding is that
* `resolveSegment(undefined)`
* `resolveSegment(null)`

will be accepted if `--structNullChecks` is false, but `resolveSegment()` will not; if `--strictNullChecks` is true, then they are all errors.

If the type is `segment?: Segment | Subsegment | null`, then they're all OK, regardless of `--strictNullChecks`, which I think is what we want?

Would you accept 2.x patches for this sort of thing? If so I'm happy to open a PR.

willarmiros:
That makes sense. I agree that we want your proposal, and I'd accept a PR but our release process is quite cumbersome so it might only end up getting released as 3.0. If it's clearly a large pain for customers, I'll also release it as 2.5.1.

esilkensen:
Sounds good, thanks -- yeah this change doesn't feel like it warrants a new release by itself. I'll open a PR!

LosD:
(moved from the issue, as this seems like a better place. If you want me to open a proper issue, I will. But I think it's just a question for now)

@willarmiros @esilkensen Is it on purpose that @types/continuation-local-storage has been added to dependencies instead of dev-dependencies? That seems like a mistake, as it'll make the package larger (right now, that of course drowns in comparison to the aws-sdk, but when that is removed, it's the largest dependency by far)

esilkensen:
Hey @LosD, since the `@types/cls-hooked` package (or `@types/continuation-local-storage` as it is now) is referenced as part of this package's public API, TypeScript users will have to download those types one way or the other, otherwise the compiler won't be able to resolve that `import { Namespace } from 'cls-hooked';` statement in context_utils.d.ts.

I think it's most convenient as a `dependency`, but it could be listed as a `peerDependency`, prompting _all_ users of the package to install the typings separately, which may not be ideal for non-TypeScript users?

Anyway, it looks like the `@types/cls-hooked` and `@types/continuation-local-storage` packages are only about 4KB each -- maybe you were looking at the size of the `continuation-local-storage` dependency itself (100KB)?

LosD:
@esilkensen Hmmm, you may be right, I mainly do application development, so maybe library development HAS to do it this way, and rely on tree-shaking not to bring it to runtime (something we currently can't do in our project, because annoying reasons :man_facepalming:).

However, while the size of `@types/continuation-local-storage` in itself is only 5.4k, it draws in `@types/node`, making the total size more than 670k (before compression, of course):
image

LosD:
Just read TypeScript's Publishing document, and they recommend doing it exactly as you are doing it, so it's just me missing finer points in library publishing. :blush:

I'll try to find out why it ends up in my runtime dependencies (other unused modules doesn't seem to)

esilkensen:
Ah, thanks, good call -- yes, and`@types/cls-hooked` has the same dependency on `@types/node`.

Well, I'm interested what others think about this, too...

Since this package is _only_ referencing that single `Namespace` interface from the `@types` dependency, it seems like if the size is prohibitive it might not be an issue to even remove the dependency altogether and just mark the return type with `any` or some approximation of the real interface.

esilkensen:
Or going back to your initial point about not needing the types at runtime, maybe it gets to (like others mentioned up above) whether we do want the types packaged with this library or if they're better off as a separate `@types` dependency, because of the package size. In any case, if there is anything I can do to help out I will be happy to.

LosD:
For me, I kinda like that it's integrated; It shows that the project takes responsibility (though in some cases, that's a promise that's not kept), but if it packing developer-only stuff into users' runtimes, that might be a good reason to avoid it. But then again, it is probably me that needs to get the project packaging fixed.

Issue #113, opened 2019-04-15
Can the bundle size be reduced?


mrchief:
According to [bundlephobia](https://bundlephobia.com/result?p=aws-xray-sdk-core@2.2.0), the core package weighs at 98kb minified+gzipped. Is there any way to reduce this bundle size?

![image](https://user-images.githubusercontent.com/781818/56154513-af21e000-5f86-11e9-8ff4-6181b87b966c.png)

lodash and winston seem to be obvious targets. lodash supports tree shaking (as well as has individual packages that are lighter). winston is a fatty package in itself with a lot of lighter & faster replacements.

I think it'd help to bring the size down especially for AWS Lambda packages which themselves can be very lightweight. 

Thoughts?

chrisradek:
@mrchief 
#99 was just merged to remove the `lodash` dependency. We have to be more careful about replacing `winston` as we would need a logger that output the same way `winston` does today but otherwise I think that's a fair change to make.

Longer-term we'll also be able to replace the `aws-sdk` with [@aws-sdk/client-xray-node](https://github.com/aws/aws-sdk-js-v3/tree/master/packages/client-xray-node) which should be a lot smaller, but will want to wait until that project moves out of developer preview first.

mrchief:
> #99 was just merged to remove the lodash dependency.

@chrisradek This is great news!

> We have to be more careful about replacing winston as we would need a logger that output the same way winston does today but otherwise I think that's a fair change to make.

Yes, I totally understand. I'm not deeply familiar with the subject, so I can't say for certain whether this would work out, but I've found [pino](https://getpino.io/#/) to be very modular (in addition to being fast and lightweight). It has a big [ecosystem](https://getpino.io/#/docs/ecosystem) of modules so you can pick and choose the ones you actually need.

Ultimately, I'd leave up to the maintainers to make the call but I'm happy that things are proceeding in the right direction!

jplock:
When can a release be made that moves the `aws-sdk` dependency to a peerDependency?

willarmiros:
Hi @jplock,

I explained in #187 why making `aws-sdk` a peer dependency is not a trivial matter. We would like to create a separate module for centralized sampling, which consumes `aws-sdk`, so it does not need to be imported by default. However, the way the package is currently designed makes sampling logic tightly coupled with many other parts of the X-Ray SDK, so making such a separate module is not an easy task. For that reason I cannot give a solid timeline on when such a release will be made.

[...]

PR #229, opened 2019-12-06
Remove date-fns dependency

willarmiros:
*Issue #, if available:*
#113 

*Description of changes:*
Removes dependency on `date-fns`. It was only used to create timestamps for our logger, a task that I implemented instead in native JavaScript. Customers complained that the dependency, when non-tree-shaken, was several MB large.

Tested locally by comparing log statements before and after and confirmed timestamps are formatted identically. Lambda environments unaffected since we don't print timestamps there.

### DIFF
--- a/packages/core/lib/logger.js
+++ b/packages/core/lib/logger.js
@@ -35,20 +35,19 @@ function calculateLogLevel(level) {
   return defaultLogLevel;
 }
 
-function createTimestamp() {
-  var now = new Date(),
-      tzo = -now.getTimezoneOffset(),  // Negate to make this tzo = local - UTC
+function createTimestamp(date) {
+  var tzo = -date.getTimezoneOffset(),  // Negate to make this tzo = local - UTC
       dif = tzo >= 0 ? '+' : '-',
       pad = function(num) {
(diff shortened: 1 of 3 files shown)


packages/core/lib/logger.js
```diff
@@ -38,7 +36,19 @@ function calculateLogLevel(level) {
...
+  var now = new Date(),
+      tzo = -now.getTimezoneOffset(),  // Negate to make this tzo = local - UTC
+      dif = tzo >= 0 ? '+' : '-',
+      pad = function(num) {
+          var norm = Math.floor(Math.abs(num));
+          return (norm < 10 ? '0' : '') + norm;
+      };
+
+  return new Date(now.getTime() + (tzo * 60 * 1000)).toISOString()
+    .replace(/T/, ' ')
```
shengxil: Shall we use 4 space indent for all lines?

willarmiros: Good call.

willarmiros:
@shengxil Thanks for the comments. I added some unit tests for the timestamp function and fixed the spacing.

PR #234, opened 2020-01-02
Add more validation to Subsegment.addAnnotation to match documentation.

joshgoodwin:
*Issue #, if available:*
An issue was posted for the Python library, and I had run into it in the Node library.

*Description of changes:*
Added validation to the addAnnotation method within the subsegment.js. Added tests to segment.test.js. Made sure to check string length properly for annotation values which can have unicode characters. 

### DIFF
--- a/packages/core/lib/segments/attributes/subsegment.js
+++ b/packages/core/lib/segments/attributes/subsegment.js
@@ -112,8 +112,8 @@ Subsegment.prototype.addPrecursorId = function(id) {
 /**
  * Adds a key-value pair that can be queryable through GetTraceSummaries.
  * Only acceptable types are string, float/int and boolean.
- * @param {string} key - The name of key to add.
- * @param {boolean|string|number} value - The value to add for the given key.
+ * @param {string} key - The name of key to add. Alphanumeric and underscores allowed. 500 characters allowed.
+ * @param {boolean|string|number} value - The value to add for the given key. If string, must be less than 1000 unicode characters.
  */
 
 Subsegment.prototype.addAnnotation = function(key, value) {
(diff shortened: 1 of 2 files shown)


joshgoodwin:
The aforementioned documentation:

https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-nodejs-segment.html#xray-sdk-nodejs-segment-annotations

packages/core/lib/segments/attributes/subsegment.js
```diff
@@ -123,6 +123,15 @@ Subsegment.prototype.addAnnotation = function(key, value) {
   } else if (typeof key !== 'string') {
     throw new Error('Failed to add annotation key: ' + key + ' value: ' + value + ' to subsegment ' +
       this.name + '. Key must be of type string.');
+  } else if (typeof key === 'string'  && key.length > 500) {
+    throw new Error('Failed to add annotation key: ' + key + ' value: ' + value + ' to subsegment ' +
+      this.name + '. Key must not be longer than 500 characters.');
+  } else if (typeof key === 'string' && !/^[A-Za-z0-9_]+$/.test(key)) {
+    throw new Error('Failed to add annotation key: ' + key + ' value: ' + value + ' to subsegment ' +
+      this.name + '. Key must be alphanumeric or underscores.');
```
willarmiros: We generally do not do any regex checking on the SDK side due to performance. Based on some quick benchmarking this change could slow down adding an annotation anywhere from 7-90 times based on the length of the key. Also, similar to above, please log a warning and drop the annotation rather than throwing a runtime error if the key doesn't comply. 

An alternative is to try to match the key against an array of acceptable characters as is done in the [Python SDK](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/models/entity.py#L142). If this can be done in a succinct way that doesn't substantially impact performance, I'm open to a revision.

packages/core/lib/segments/attributes/subsegment.js
```diff
@@ -123,6 +123,15 @@ Subsegment.prototype.addAnnotation = function(key, value) {
   } else if (typeof key !== 'string') {
     throw new Error('Failed to add annotation key: ' + key + ' value: ' + value + ' to subsegment ' +
       this.name + '. Key must be of type string.');
+  } else if (typeof key === 'string'  && key.length > 500) {
```
willarmiros: We don't check the length of keys and values in the other SDKs because they are truncated to the appropriate length on the X-Ray backend. We would not want to start throwing runtime errors for annotations in one language that are valid in another.

packages/core/test/unit/segments/segment.test.js
```diff
@@ -113,6 +113,32 @@ describe('Segment', function() {
...
+      value = 'value';
+    });
+
+    it('should throw an error if trying to add with invalid key', function() {
+      assert.throws( function() { subsegment.addAnnotation('invalid key with non-alphanumerics', 'validValue'); }, Error);
+      assert.throws( function() { subsegment.addAnnotation('invalid_key_because_of_length'+'1'.repeat(500), 'validValue'); }, Error);
+      assert.throws( function() { subsegment.addAnnotation(3, 'validValue'); }, Error);
+    });
+
+    it('should throw an error if trying to add with invalid value', function() {
```
willarmiros: Fix typo in description.

Issue #240, opened 2020-01-15
Got 10 doesn't work with AWS XRay in lambda


MickVanDuijn:
#### Describe the bug

 - Node.js version: 10.x and 12.x
 - OS & version: AWS Lambda, Got 10

This occurs when using Got 10 in combination with AWS XRay on AWS Lambda. **Note**: Got 9 works fine with the provided snippet. 

I also opened an issue on the [`got`](https://github.com/sindresorhus/got/issues/1022) repository.

#### Actual behavior

Whenever XRay capture is enabled, the lambda doesn't return any response, and the actual request is made against `https://localhost`. Subsequent executions take longer and longer (even up to 2 minutes) until finally one execution will show a bunch of LAMBDA_RUNTIME_ERROR. 

**Edit**: the lambda execution actually dies midway, while without XRay enabled, it successfully completes.

#### Expected behavior

Get contents of `google.com` and return the specified response (`body: 'ok'`)

#### Code to reproduce

```js
require('aws-xray-sdk-core').captureHTTPsGlobal(require('https'))
const got = require('got')

export async function handler() {
    await got.get(`https://google.com`)

    return {
        statusCode: 200,
        body: 'ok',
    }
}
```

<img width="1171" alt="image" src="https://user-images.githubusercontent.com/1350647/72446204-6320f780-37b3-11ea-956e-fc53477c3d1e.png">

MickVanDuijn:
@chanchiem should this also receive the `bug` label? It is killing my executions midway, while it worked fine without XRay enabled

willarmiros:
Hi @MickVanDuijn,
Thank you for raising this issue. I was able to reproduce it and the specific fault I got on the X-Ray console was:
```
Error: connect ECONNREFUSED 127.0.0.1:443
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1128)
```

That makes me think this issue might be related to the one in #217, and it seems like that host implies the HTTP request can't even get out to the internet. I'm investigating that issue with Lambda, and I'll add this issue to that investigation.

MickVanDuijn:
Thanks @willarmiros ! What's also strange is that the requested URL turned into `localhost`, I don't see that on #217

willarmiros:
Hi @MickVanDuijn,
[...]

PR #247, opened 2020-01-31
Support new syntax for https.request

MickVanDuijn:
**Issue #240**

*Description of changes:*
Support the new `request(url, options, callback)` syntax from NodeJS >= `v10`: https://nodejs.org/docs/latest-v10.x/api/https.html#https_https_request_url_options_callback

### DIFF
--- a/packages/core/lib/patchers/http_p.js
+++ b/packages/core/lib/patchers/http_p.js
@@ -59,20 +59,22 @@ function enableCapture(module, downstreamXRayEnabled) {
     let callback;
     let hasUrl;
     let urlObj;
+
+    let arg0 = args[0];
     if (typeof args[1] === 'object') {
       hasUrl = true;
-      urlObj = typeof args[0] === 'string' ? url.parse(args[0]) : args[0];
+      urlObj = typeof arg0 === 'string' ? url.parse(arg0) : arg0;
       options = args[1],
       callback = args[2];
     } else {
       hasUrl = false;
-      options = args[0];
+      options = arg0;
       callback = args[1];
     }
+
     if (!options || (options.headers && (options.headers['X-Amzn-Trace-Id']))) {
       return baseFunc(...args);
     }
-
     if (typeof options === 'string') {
       options = url.parse(options);
     }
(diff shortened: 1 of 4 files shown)


willarmiros:
Thank you very much for the contribution @MickVanDuijn! I'll need some time to review it since there are many backlog items right now, but I appreciate the move for forwards-compatibility.

packages/core/lib/patchers/http_p.js
```diff
@@ -113,7 +131,7 @@ function enableCapture(module, downstreamXRayEnabled) {
 
     var optionsCopy = Utils.objectWithoutProperties(options, ['Segment'], true);
 
-    var req = baseFunc(optionsCopy, function(res) {
+    var req = baseFunc(...(hasUrl ? [urlObj, optionsCopy] : [options]), function(res) {
```
willarmiros: You must not pass the `urlObj` here, instead pass the `url` string. I just passed in `args[0]` to verify it works, but I'd suggest saving the URL string into a variable above.

MickVanDuijn: thanks! will update my PR asap!

MickVanDuijn: @willarmiros I modified the code to pass along `args[0]` unmodified in case of the new syntax (simply named it `let arg0 = args[0]` to prevent assumptions about the type, and to reuse it in several places)

MickVanDuijn:
Thanks! Then I'll eagerly await the next release 🚀

Mangosteen-Yang:
Just want to know is this has any improvement for axios?

MickVanDuijn:
Possibly yes, if Axios also uses the new `request` syntax with more parameters, then definitively yes!

rictorres:
@willarmiros is it possible to publish a release with this fix to npm?

willarmiros:
@rictorres this change is released in `aws-xray-sdk-core@3.0.0-alpha.2`. See [CHANGELOG](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/CHANGELOG.md).

If you need a stable version you can track #157 for the release of 3.0.0, which will be the next release I make.

Issue #244, opened 2020-01-28
MySQL query timeout gets ignored


MMulaosmanovic:
When using captureMySQL it seems to be stripping out all of the options passed to Query, line 200 of the following seems to only pass the sql property.

https://github.com/aws/aws-xray-sdk-node/blob/master/packages/mysql/lib/mysql_p.js 

Is there any way to use X-Ray for MySQL and still pass options to the query like a timeout like this?

https://github.com/mysqljs/mysql#performing-queries 
`
connection.query({
  sql: 'SELECT * FROM `books` WHERE `author` = ?',
  timeout: 40000, // Gets ignored because xray is not explicitly passing it 
  values: ['David']
}, function (error, results, fields) {
});
`

willarmiros:
Hi @MMulaosmanovic,
Thanks for raising this issue. Because of the way that the `mysql` library is monkey patched, it may not be forwards-compatible with parameters added after the implementation of the patcher. At this time it is not possible to record the timeout parameter because it is filtered out in the [resolveArguments](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/mysql/lib/mysql_p.js#L138) method. 

We are always open to PRs if you'd like to fix this issue yourself. I've added it to our backlog as well.

MMulaosmanovic:
Would you be open to just adding the timeout property?

If so I can submit this pull request  https://github.com/shipworks/aws-xray-sdk-node

willarmiros:
Sure thing, feel free to open it and I'll review it.

willarmiros:
Released in v3.0.0-alpha.2

PR #248, opened 2020-01-31
Support MySQL timeout 

MMulaosmanovic:
https://github.com/aws/aws-xray-sdk-node/issues/244

use timeout if it is present 


willarmiros:
Hi @MMulaosmanovic,
Thank you for this PR, I'll take a look as soon as I can. For future reference it does not appear to need any change to the `mysql_p.d.ts` file since the `timeout` parameter is a field of `MySQL.QueryOptions` ([reference](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/mysql/lib/mysql_p.d.ts#L9)).

Issue #267, opened 2020-03-26
3.0.0-alpha.2 - attempting to fetch sampling rules crashes application when agent is not running


pdhodgkinson:
We've just upgraded our application from alpha.1 to alpha.2 for testing and immediately noticed that the application exited in cases where the xRay agent was not running on the configured url:port.
This was not the behaviour in alpha.1

The application is an express server, and we are using the `express_mw` module, and using `AWSXRay.captureHTTPsGlobal`.

I did some digging and found the difference appears to be somewhere in the hierarchy that starts with `mwUtils.resolveSampling(amznTraceHeader, segment, res);`
https://github.com/aws/aws-xray-sdk-node/blob/aws-xray-sdk%403.0.0-alpha.2/packages/express/lib/express_mw.js#L40
and proceeds down to `service_connector.js:fetchSamplingRules`
https://github.com/aws/aws-xray-sdk-node/blob/aws-xray-sdk%403.0.0-alpha.2/packages/core/lib/middleware/sampling/service_connector.js#L23

Here there appears to be some custom http request handling, that, when there is no agent running, or it's running on a different port,  throws an `ECONNREFUSED` error like
`
  trace: [
    {
      column: 16,
      file: 'net.js',
      function: 'TCPConnectWrap.afterConnect [as oncomplete]',
      line: 1134,
      method: 'afterConnect [as oncomplete]',
      native: false
    }
  ],
  stack: [
    'Error: connect ECONNREFUSED 127.0.0.1:2000',
    '    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1134:16)'
  ]`
This does not get caught anywhere in the stack, and results in a runtime `uncaughtException` that terminates the application. It seems to sidestep any express error capturing logic I put in.

In `alpha.1` it appears that fetchSamplingRules was implemented using a completely different http client (the AWS-SDK request library?), and for some reason that seems to have swallowed and hidden any ECONNREFUSED errors.
https://github.com/aws/aws-xray-sdk-node/blob/aws-xray-sdk%403.0.0-alpha.1/packages/core/lib/middleware/sampling/service_connector.js#L22

An ideal fix (for us) would be to capture and log any errors connecting (or using) the agent to the logger and not crash the application. Or possibly make the error behaviour configurable? Note that we are already using the ` AWSXRay.setContextMissingStrategy` to avoid the RUNTIME_ERROR cases for missing context.

willarmiros:
[...]

PR #268, opened 2020-03-26
Catch errors on request to daemon

willarmiros:
*Issue #, if available:*
#267

*Description of changes:*
Fixes a bug where a runtime error is thrown if the `service_connector` cannot connect to the daemon because it is not present. We log the error instead. Also added unit test to check the condition.

Added some additional logging changes to make statements more clear and consistent between rule & target poller.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "aws-xray-sdk-node",
-  "version": "3.0.0-alpha.1",
+  "version": "3.0.0-alpha.2",
   "private": true,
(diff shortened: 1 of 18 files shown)


ianmetcalf:
Related typo: https://github.com/aws/aws-xray-sdk-node/blob/634386e3725ffe0e2341869782e68653f384a9c1/packages/core/lib/middleware/sampling/service_connector.js#L42-L44

willarmiros:
@ianmetcalf good catch thanks! Fixed.

PR #284, opened 2020-04-24
Fix trace header behavior

willarmiros:
*Description of changes:*
* Added TraceID class
* Made trace header parsing case-insensitive
* Added trace ID validation to header parsing
* Ensured that Lambda will never use a `null` trace ID in downstream headers
* Added unit tests

Tested changes on Lambda & express sample apps.

### DIFF
--- a/packages/core/lib/utils.js
+++ b/packages/core/lib/utils.js
@@ -2,7 +2,9 @@
  * @module utils
  */
 
+var crypto = require('crypto');
 var logger = require('./logger');
+var TraceID = require('./segments/attributes/trace_id');
 
 var utils = {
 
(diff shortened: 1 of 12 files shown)


packages/core/lib/utils.js
```diff
@@ -136,33 +138,46 @@ var utils = {
...
+      } else if (!data.root || data.root !== TraceID.FromString(data.root).toString() 
+                 || !data.parent || !data.sampled) 
+      {
+        logger.getLogger().error('_X_AMZN_TRACE_ID is missing required information');
+      } else {
+        valid = true;
+      }
 
-        logger.getLogger().debug('Segment started: ' + JSON.stringify(data));
+      segment.trace_id = TraceID.FromString(data.root).toString();  // Will always assign valid trace_id
```
shengxil, line 171: `TraceID.FromString(data.root).toString()` was called twice, which can be a little bit expensive Can we merge them into one call?

packages/core/lib/utils.js
```diff
@@ -136,33 +138,46 @@ var utils = {
...
+      segment.id = data.parent || crypto.randomBytes(8).toString('hex');
 
-        populated = true;
-      } else
-        logger.getLogger().warn('_X_AMZN_TRACE_ID is missing required data.');
+      if (!parseInt(data.sampled))
```
shengxil, line 174: It looks like `parseInt` returns a 0 if the input is "?" which is a valid sampling decision ("Requested"). https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

Can we create a SamplingDecision enum like Java?

packages/core/test/unit/segments/attributes/trace_id.test.js
```diff
@@ -0,0 +1,47 @@
+var assert = require('chai').assert;
```
shengxil, line 1: Can we create a test to verify trace header keys are case insensitive?

packages/core/lib/utils.js
willarmiros: Good call, I'll only use it once.

packages/core/test/unit/segments/attributes/trace_id.test.js
willarmiros: Already have one [here](https://github.com/aws/aws-xray-sdk-node/pull/284/files/76671cbfa5ae532306a2c29820b67ac91b48cf4c#diff-6a36c6c7818cffd493bc93a85aebe9b4R33-R40) :)

packages/core/lib/utils.js
willarmiros: Yes, currently `?` is handled the same way as a 0 sampling decision because currently in the node SDK the sampling decision is modeled as strictly a boolean. While I agree it should be an enum eventually, changing the segment model to use an `enum` instead of a `boolean` would be a pretty big undertaking for this PR. What do you think about the priority?

willarmiros:
@shengxil I did make the trace header parsing case-insensitive, see [here](https://github.com/aws/aws-xray-sdk-node/pull/284/files/76671cbfa5ae532306a2c29820b67ac91b48cf4c#diff-f1162832840b06f003c3d317708874c4R204).

packages/core/test/unit/segments/attributes/trace_id.test.js
```diff
@@ -0,0 +1,47 @@
...
+var TraceID = require('../../../../lib/segments/attributes/trace_id');
+
+function validateTraceID(traceID) {
```
bhautikpip, line 4: Can we write a test when there is overlong arbitrary field in traceheader  ? Ex. Root=1-58406520-a006649127e371903a2de979;Parent=53995c3f42cd8ad8;Sampled=1;Foo=Bar;[A field bigger than 256 bytes]

willarmiros: @bhautikpip Good call. I added this functionality & tests [here](https://github.com/aws/aws-xray-sdk-node/pull/284/files#diff-6a36c6c7818cffd493bc93a85aebe9b4R92-R111)!

packages/core/lib/utils.js
shengxil: Yeah it doesn't have to be in this PR but I'm afraid it becomes another time bomb like the "dummy" trace id if we lost track of it. There's a big disagreement:`?` means "to be decided" while 0 means "not sampled". Shall we create an issue on fixing this?

packages/core/test/unit/middleware/mw_utils.test.js
```diff
@@ -200,7 +200,7 @@ describe('Middleware utils', function() {
     });
 
     it('should set the response header with sampling result if header is "?"', function() {
-      var headers = { Root: traceId, Sampled: '?' };
+      var headers = { root: traceId, sampled: '?' };
       MWUtils.resolveSampling(headers, segment, res);
 
       var expected = new RegExp('^Root=' + traceId + ';Sampled=1$');
```
shengxil, line 206: This expectation should be corrected too later

packages/core/lib/utils.js
willarmiros: Totally agree. I've made #291 to track it.

Issue #285, opened 2020-05-02
Application crash on missing data['SamplingTargetDocuments']


JeremieDoctrine:
Hello,

I'm using the `aws-xray-sdk-core: 3.0.0` and when I turn it on on production, I sometimes have the following error.

```
TypeError: Cannot read property 'forEach' of undefined
at assembleTargets (/app/node_modules/aws-xray-sdk-core/lib/middleware/sampling/service_connector.js line 147 col 7)
docs.forEach(function(doc) {
at IncomingMessage.res.on (/app/node_modules/aws-xray-sdk-core/lib/middleware/sampling/service_connector.js line 82 col 29)
var targetsMapping = assembleTargets(dataObj);
at IncomingMessage.emit (events.js line 203 col 14)
at IncomingMessage.EventEmitter.emit (domain.js line 448 col 19)
at endReadableNT (_stream_readable.js line 1145 col 11)
at process._tickCallback (internal/process/next_tick.js line 63 col 18)
```

Looks like there is an issue between the SDK and the daemon, but not sure where exactly.
One thing to know is that I've set all my rules to `Limit to 0 r/sec, then 0% fixed rate`

willarmiros:
Hmmm you appear to be getting an invalid response from the X-Ray service. The SDK polls the service roughly every ten seconds using the [GetSamplingTargets](https://docs.aws.amazon.com/xray/latest/api/API_GetSamplingTargets.html#API_GetSamplingTargets_ResponseSyntax) API to ensure the SDK is performing sampling at an appropriate rate. The error you're getting seems to imply there is no `SamplingTargetDocuments` in the response. 

How often are you receiving this error? If it's rare, then it's probably just an issue with the response occasionally getting jumbled over the network and we should be able to resolve this by doing better validation of the response.

JeremieDoctrine:
I turned it on on production for one day and received this error 7 times for 8 containers. So I guess it's rare enough.

What's worrying me is that I'm the only one reporting it, there is probably something wrong with my network as well.

willarmiros:
Relative to the frequency of the polling that is fairly rare. To confirm, the sampling rate you set is being read correctly right? Like you've seen no segments being emitted to the X-Ray console as you'd expect with your sampling rate?

JeremieDoctrine:
Yes, the sampling rate I set is being read correctly.

[...]

PR #287, opened 2020-05-04
Improved API response validation

willarmiros:
*Issue #, if available:*
#285 

*Description of changes:*
Adds additional validation logic and unit tests to make sure the SDK gracefully (non-disruptively) handles invalid responses from the sampling APIs. 

### DIFF
--- a/packages/core/lib/middleware/sampling/service_connector.js
+++ b/packages/core/lib/middleware/sampling/service_connector.js
@@ -43,6 +43,10 @@ var ServiceConnector = {
           callback(err);
         }
 
+        if (dataObj === null) {
+          callback(new Error('AWS X-Ray GetSamplingRules API returned null response'));
+        }
+
         var newRules = assembleRules(dataObj);
         callback(null, newRules);
       });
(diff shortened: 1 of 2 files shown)


rysi3k:
Hello @willarmiros @bhautikpip 
I have problem with this PR after update to 3.0.1 of sdk.
In logs I see many errors `Error: AWS X-Ray SamplingTargets API returned invalid response`
After debugging I see that response from daemon looks like that:
```
{
  LastRuleModification: 0,
  SamplingTargetDocuments: [
    {
      FixedRate: 0.05,
      Interval: 10,
      ReservoirQuota: 0,
      ReservoirQuotaTTL: 1601894236,
      RuleName: 'Default'
    }
  ],
  UnprocessedStatistics: []
}
```
It looks ok right? So why in code is condition to throw if `LastRuleModification` is 0 or unset?
Thanks for help.

willarmiros:
@rysi3k that was a miss on my end, it should treat that response as valid and is only creating an error because the value 0 is falsey. I'll get a fix out.

PR #297, opened 2020-05-22
Add support for IMDSv2

willarmiros:
*Description of changes:*
Changed the EC2Plugin to use instance metadata service (IMDS) v2 first to collect instance metadata, and use IMDSv1 as a fallback. One difference between this implementation and the [python](https://github.com/aws/aws-xray-sdk-python/pull/226) is that we retrieve the [dynamic instance identity document](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html) instead of calling the instance ID and AZ endpoints individually. This is because the Node SDK retrieves the document asynchronously, so even though its requests take longer the customer won't notice.

There were some issues with the original `Plugin.js` implementation, which is meant to be a helper class for making metadata endpoint requests, that I addressed. Specifically, I reduced the number of retries from 20 to 5, since that was very excessive. I also changed the retry logic to retry on 5xx errors instead of 4xx errors, since 4xx errors will persist irrespective of how many retries you make.

Tested this on an EC2 instance with V2 metadata enabled.

### DIFF
--- a/packages/core/lib/segments/plugins/ec2_plugin.js
+++ b/packages/core/lib/segments/plugins/ec2_plugin.js
@@ -9,15 +9,18 @@ var EC2Plugin = {
    */
 
   getData: function(callback) {
+    const TOKEN_PATH = '/latest/api/token';
+    const METADATA_PATH = '/latest/dynamic/instance-identity/document';
+
     function getMetadataToken() {
       const TTL = 60; //seconds
-      const tokenOptions = getOptions('/api/token', 'PUT', {
+      const tokenOptions = getOptions(TOKEN_PATH, 'PUT', {
         'X-aws-ec2-metadata-token-ttl-seconds': TTL
       });
 
       Plugin.getToken(tokenOptions, function(err, data) {
         if (err) {
-          logger.getLogger().debug('EC2Plugin failed to get token from IMDSv2. Falling back to IMDSv1.', err);
+          logger.getLogger().debug('EC2Plugin failed to get token from IMDSv2. Falling back to IMDSv1.', err.toString());
           getV1Metadata();
         } else {
           // Successfully got token, use it for IMDSv2 metadata request
@@ -27,17 +30,20 @@ var EC2Plugin = {
     }
 
     function getV2Metadata(token) {
-      const options = getOptions('/latest/dynamic/instance-identity/document', 'GET', {
+      const options = getOptions(METADATA_PATH, 'GET', {
         'X-aws-ec2-metadata-token': token
       });
 
       Plugin.getPluginMetadata(options, function(err, data) {
         if (err) {
-          logger.getLogger().debug('EC2Plugin failed to get metadata from IMDSv2. Falling back to IMDSv1.', err)
+          logger.getLogger().debug('EC2Plugin failed to get metadata from IMDSv2. Falling back to IMDSv1.', err.toString())
           getV1Metadata();
-        } else {
+        } else if (data) {
           const metadata = { ec2: { instance_id: data.instanceId, availability_zone: data.availabilityZone }};
           callback(metadata);
+        } else {
+          logger.getLogger().debug('Failed to get required EC2 metadata from IMDSv2. Falling back to IMDSv1.');
+          getV1Metadata();
         }
       });
     }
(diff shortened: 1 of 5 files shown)


packages/core/lib/segments/plugins/plugin.js
```diff
@@ -17,30 +24,75 @@ var Plugin = {
...
 
+      req.setTimeout(Plugin.METADATA_TIMEOUT);
       req.end();
     };
 
     getMetadata();
+  },
+
+  /**
+   * Asynchronously retrieves a token used in requests to EC2 instance metadata service.
```
anuraaga, line 63: I think this method is passing `option` to the request but doesn't really have any logic specific to retrieving the token. How about removing the `options` callback and setting up the request here? The options look like constants that can be inlined here and then this method does become a `getToken` method :)

packages/core/lib/segments/plugins/ec2_plugin.js
```diff
@@ -9,22 +9,86 @@ var EC2Plugin = {
...
+          const metadata = { ec2: { instance_id: data.instanceId, availability_zone: data.availabilityZone }};
+          callback(metadata);
+        } else {
+          logger.getLogger().debug('Failed to get required EC2 metadata from IMDSv2. Falling back to IMDSv1.');
+          getV1Metadata();
+        }
+      });
+    }
+
+    function getV1Metadata() {
```
anuraaga, line 51: As far as I can tell, the only difference between v1 and v2 is whether the token is included or not. Can we share the business logic, mostly the URL and the response parsing?

anuraaga: Also makes me wonder, is it actually possible for the request to fail? I guess for IDMS v1 it'll just ignore the header? It doesn't hurt to have the fallback logic but it'd be nice to understand what the expected behavior in prod is.

anuraaga: Just in case it helps, here's some code in Java I wrote that has less branching. Let me know if it's broken though ;)

https://github.com/aws-samples/aws-xray-sdk-with-opentelemetry-sample/blob/both-otel-and-xray/awsagentprovider/src/main/java/com/softwareaws/xray/opentelemetry/exporters/resources/Ec2ResourcePopulator.java#L64

packages/core/lib/segments/plugins/ec2_plugin.js
```diff
@@ -9,22 +9,86 @@ var EC2Plugin = {
...
+    }
+
+    function getV2Metadata(token) {
+      const options = getOptions(METADATA_PATH, 'GET', {
+        'X-aws-ec2-metadata-token': token
+      });
+
+      Plugin.getPluginMetadata(options, function(err, data) {
+        if (err) {
+          logger.getLogger().debug('EC2Plugin failed to get metadata from IMDSv2. Falling back to IMDSv1.', err.toString())
```
srprash, line 39: nit: Missing ; at the end

packages/core/lib/segments/plugins/plugin.js
srprash: I agree with anuraag here that this function doesn't really do anything specific to token. Also, token is currently only used with IMDSv2 which is only available for EC2, so if you're considering to make this function into a real `getToken` function, maybe moving it to `ec2_plugin.js` would make more sense.

packages/core/lib/segments/plugins/plugin.js
```diff
@@ -1,10 +1,17 @@
 var http = require('http');
```
srprash, line 1: if a customer has enabled tracing of all http clients by doing `AWSXRay.captureHTTPsGlobal(require('http'))`, it would trace the calls to the metadata service too right?

willarmiros: This is handled later on by:
```js
var httpReq = http.__request ? http.__request : http.request;
```

Which ensures we're using the non-instrumented HTTP client.

packages/core/lib/segments/plugins/plugin.js
willarmiros: Both fair callouts, I'll move it to `ec2_plugin` and narrow its scope.

packages/core/lib/segments/plugins/ec2_plugin.js
willarmiros: That's a good observation. Really only the token request would fail if we were on a V1-only instance. So I'll refactor to share much more of the logic for the two versions based on your sample. That'll probably make its way to the Java SDK as well :)

Issue #294, opened 2020-05-12
Support for AWS SDK for JavaScript v3


anvar:
Is there a branch parked somewhere with support for the new [JavaScript AWS SDK](https://github.com/aws/aws-sdk-js-v3)? The new SDK uses middleware to intercept and transform requests/responses, which is a bit different to how the xray-sdk hooks into the 2.x version. If not, happy to get started and send a PR, just thought I would check beforehand.

Thanks!

willarmiros:
Hi @anvar, 
We are not yet developing support for AWS SDK for JS v3. We'd be open to a pull request if you're interested. The only caveats are that we'd like to keep dependencies to an absolute minimum, although that shouldn't be an issue with v3's modular design. We'll also need to discuss with the AWS SDK team about the stability of the middleware APIs before we can mainline our own middleware, since they are still in beta.

Otherwise, contribute away!

anvar:
Great, thank you, I will have a look!

trivikr:
Update: The [general availability of the AWS SDK for JavaScript, version 3 (v3)](https://aws.amazon.com/blogs/developer/modular-aws-sdk-for-javascript-is-now-generally-available/) was announced on December 15th, 2020.

awssandra:
Reviewing pull request in flight.

nickitza:
Hey there, saw this was merged to master, but am getting the error "AWSXRay.captureAWSv3Client is not a function" after implementing the S3 client module with AWSXRay in a Lambda. Does version 3.2.0 contain these changes? Thanks!

bfaulk96:
Hey all, I noticed https://github.com/aws/aws-xray-sdk-node/pull/386 got merged, but it seems I still get the `TypeError: service.customizeRequests is not a function` error when trying to capture the Dynamo Document Client from the sdk v3... Any updates?

willarmiros:
@bfaulk96 see #395

willarmiros:
Released in 3.3.0!

willarmiros:
Reopening because we had to temporarily rollback AWS SDK V3 support in patch release v3.3.1  to address #411. We are actively working on another patch that will add back AWS SDK v3 support in 3.3.2.

bfaulk96:
Any updates on this being re-implemented?

awssandra:
We are actively addressing this to re-release. Stay tuned!

jwoehrle:
[...]

PR #298, opened 2020-05-27
refs #294, add support for AWS SDK v3

monken:
#294

* add support for AWS SDK v3

I used typescript for the actual patcher because it makes the code safer and I wanted to make sure that types are maintained when using the XRay patcher. I included the complied JavaScript code as well so that tests can be run without adding more typescript flavor to the project.
There is no `captureAWS` for now and I also haven't looked at manual mode much because I'm unsure about how to pass in the xray params in a type-safe way.

Obviously the aws-sdk-v3 is not GA yet so this just serves as a placeholder and to get some additional eyes on the code.

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -464,6 +464,8 @@ function sendRequest(host, cb) {
 
 ### Capture all outgoing AWS requests
 
+This is only available for AWS SDK v2 due to the service-oriented architecture of AWS SDK v3.
+
 ```js
 var AWS = captureAWS(require('aws-sdk'));
 
@@ -471,6 +473,23 @@ var AWS = captureAWS(require('aws-sdk'));
 ```
 
 ### Capture outgoing AWS requests on a single client
+
+
+AWS SDK v3
+
+```js
+import { S3, PutObjectCommand } from '@aws-sdk/client-s3';
+
+const s3 = AWSXRay.captureAWSv3Client(new S3({}));
+
+await s3.send(new PutObjectCommand({
+  Bucket: bucketName,
+  Key: keyName,
+  Body: 'Hello!',
+}));
+```
+
+AWS SDK v2
 
 ```js
 var s3 = AWSXRay.captureAWSClient(new AWS.S3());
@@ -622,6 +641,24 @@ function sendRequest(host, cb, subsegment) {
 
 ### Capture outgoing AWS requests on a single client
 
+AWS SDK v3
+
+```js
+import { S3, PutObjectCommand } from '@aws-sdk/client-s3';
+
+// subsegment is an optional parameter that is required for manual mode
+// and can be omitted in automatic mode (e.g. inside a Lambda function).
+const s3 = AWSXRay.captureAWSv3Client(new S3({}), subsegment);
+
+await s3.send(new PutObjectCommand({
+  Bucket: bucketName,
+  Key: keyName,
+  Body: 'Hello!',
+}));
+```
+
+AWS SDK v2
+
 ```js
 var s3 = AWSXRay.captureAWSClient(new AWS.S3());
 var params = {
@@ -637,6 +674,8 @@ s3.putObject(params, function(err, data) {
 ```
 
 ### Capture all outgoing AWS requests
+
+This is only available for AWS SDK v2 due to the service-oriented architecture of AWS SDK v3.
 
 ```js
 var AWS = captureAWS(require('aws-sdk'));
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -643,6 +643,9 @@ function sendRequest(host, cb, subsegment) {
 
 AWS SDK v3
 
+You must re-capture the client every time the subsegment is attached
+to a new parent.
+
 ```js
 import { S3, PutObjectCommand } from '@aws-sdk/client-s3';
 
--- a/packages/core/lib/patchers/aws3_p.ts
+++ b/packages/core/lib/patchers/aws3_p.ts
@@ -30,7 +30,7 @@ async function buildAttributesFromMetadata(
   metadata: ResponseMetadata,
 ): Promise<[ServiceSegment, HttpResponse]> {
   const { extendedRequestId, requestId, httpStatusCode: statusCode, attempts } = metadata;
-  const serviceIdentifier = client.config.signingName as string;
+  const serviceIdentifier = client.config.serviceId as string;
 
   let operation: string = command.constructor.name.slice(0, -7);
 
@@ -71,7 +71,7 @@ function addFlags(http: HttpResponse, subsegment: Subsegment, err?: SdkError): v
 }
 
 type DefaultConfiguration = RegionInputConfig & {
-  signingName: string;
+  serviceId: string;
 };
 
 export function captureAWSClient<
@@ -82,7 +82,7 @@ export function captureAWSClient<
   // create local copy so that we can later call it
   const send = client.send;
 
-  const serviceIdentifier = client.config.signingName;
+  const serviceIdentifier = client.config.serviceId;
 
   client.send = async (...args: Parameters<typeof client.send>): Promise<Output> => {
     const [command] = args;
--- a/packages/core/lib/patchers/aws3_p.ts
+++ b/packages/core/lib/patchers/aws3_p.ts
@@ -1,0 +1,173 @@
+import type {
+  MetadataBearer,
+  ResponseMetadata,
+  Command,
+  Client,
+} from '@aws-sdk/types';
+
+import type { RegionInputConfig } from '@aws-sdk/config-resolver';
+
+import { isThrottlingError } from '@aws-sdk/service-error-classification';
+
+import ServiceSegment from '../segments/attributes/aws';
+
+import { stringify } from 'querystring';
+
+import Subsegment from '../segments/attributes/subsegment';
+
+var contextUtils = require('../context_utils');
+
+var logger = require('../logger');
+
+import { getCauseTypeFromHttpStatus } from '../utils';
+import { SdkError } from '@aws-sdk/smithy-client';
+import { SegmentLike } from '../aws-xray';
+
+type HttpResponse = { response: { status: number } };
+
+async function buildAttributesFromMetadata(
+  client: any,
+  command: any,
+  metadata: ResponseMetadata,
+): Promise<[ServiceSegment, HttpResponse]> {
+  const { extendedRequestId, requestId, httpStatusCode: statusCode, attempts } = metadata;
+  const serviceIdentifier = client.config.signingName as string;
+
+  let operation: string = command.constructor.name.slice(0, -7);
+
+  const aws = new ServiceSegment(
+    {
+      extendedRequestId,
+      requestId,
+      retryCount: attempts,
+      request: {
+        operation,
+        httpRequest: {
+          region: await client.config.region(),
+          statusCode,
+        },
+        params: command.input,
+      },
+    },
+    serviceIdentifier,
+  );
+
+  const http = { response: { status: statusCode || 0 } };
+  return [aws, http];
+}
+
+function addFlags(http: HttpResponse, subsegment: Subsegment, err?: SdkError): void {
+  if (err && isThrottlingError(err)) {
+    subsegment.addThrottleFlag();
+  } else if (http.response?.status === 429) {
+    subsegment.addThrottleFlag();
+  }
+
+  const cause = getCauseTypeFromHttpStatus(http.response.status);
+  if (cause === 'fault') {
+    subsegment.addFaultFlag();
+  } else if (cause === 'error') {
+    subsegment.addErrorFlag();
+  }
+}
+
+type DefaultConfiguration = RegionInputConfig & {
+  signingName: string;
+};
+
+export function captureAWSClient<
+  Input extends object,
+  Output extends MetadataBearer,
+  Configuration extends DefaultConfiguration
+>(client: Client<Input, Output, Configuration>, manualSegment?: SegmentLike): Client<Input, Output, Configuration> {
+  // create local copy so that we can later call it
+  const send = client.send;
+
+  const serviceIdentifier = client.config.signingName;
+
+  client.send = async (...args: Parameters<typeof client.send>): Promise<Output> => {
+    const [command] = args;
+    const segment = manualSegment || contextUtils.resolveSegment();
+
+    let operation: string = command.constructor.name.slice(0, -7);
+
+    if (!segment) {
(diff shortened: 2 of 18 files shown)


monken:
@AllanFly120 it would be great to get your opinion on this implementation as well.

awssandra:
Hi monken,

We're just discussing a few changes in flight.
Hang on, we'll be right back.

Thanks!
Sandra

willarmiros:
@monken we're very sorry for the lack of transparency on this - we were waiting for the AWS SDK v3 to reach a more stable state. Are you still interested in seeing this PR merged? If so, please update your branch & resolve conflicts so the new CI tests can run and I'll give it a review.

monken:
@willarmiros sure, I can give it another go and update the pull request to the latest version if the v3 SDK. I just want to make sure that this effort is not in vain and that you are ok with general direction of this pull request.

willarmiros:
@monken my only concern is this being written in TS in an otherwise pure-JS repo, but given that AWS SDK v3 is written in TS and it doesn't cause any impact to the end user I'm ok with a bit of extra developer burden. As long as you're ok with adding the `captureAWS` feature and addressing some other (likely minor) review comments then it will not be in vain :)

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -0,0 +1,179 @@
...
+  Input extends object,
+  Output extends MetadataBearer,
+  Configuration extends DefaultConfiguration
+>(client: Client<Input, Output, Configuration>): Client<Input, Output, Configuration> {
+  // create local copy so that we can later call it
+  const send = client.send;
+
+  const serviceIdentifier = client.config.signingName;
+
+  client.send = async (command: any): Promise<Output> => {
```
willarmiros, line 94: Is there any way we could also use the `middlewareStack` for this instead of redefining the `send` operation? Totally ok if not, I'd just rather use the first-class features of the new AWS SDK where possible :)

packages/core/test/unit/patchers/aws3_p.test.js
```diff
@@ -0,0 +1,239 @@
...
+var { MiddlewareStack } = require('@aws-sdk/middleware-stack');
+
+var logger = require('../../../lib/logger').getLogger();
+
+chai.should();
+chai.use(sinonChai);
+
+var traceId = '1-57fbe041-2c7ad569f5d6ff149137be86';
+
+describe('AWS v3 patcher', function() {
```
willarmiros, line 23: In the past we've had an issue where we'd create subsegments for presigning S3 URLs, even though it's a local request. But I'm *pretty sure* this shouldn't be a problem with the new V3 design, since the S3 presigner is its [own package](https://github.com/aws/aws-sdk-js-v3/tree/master/packages/s3-request-presigner) rather than implemented as a normal S3 client operation like before. Can we add a unit test to make sure that presigns won't get captured?

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -0,0 +1,179 @@
...
+    const segment = contextUtils.resolveSegment();
```
willarmiros, line 95: Yeah, so manual mode will be tough in TypeScript. Is there any way that, when calling `captureAWSClient`, we could iterate over all its methods and add an additional optional argument for the `XRaySegment`? Or could we add it as a field to the `Configuration` object?

packages/core/test/unit/patchers/aws3_p.test.js
```diff
@@ -0,0 +1,239 @@
...
+      sandbox.stub(Aws.prototype, 'init').returns();
+
+      awsRequest.response = {
+        error: { message: 'throttling', code: 'ProvisionedThroughputExceededException' },
+        $metadata: { httpStatusCode: 400 },
+      };
+
+      await awsClient.send(awsRequest).catch(() => null);
+
+      // throttledCheckStub.should.have.been.calledOnce;
```
willarmiros, line 213: Nit: remove comment

willarmiros (commented):
@monken thanks again for this wonderful contribution! I left some comments but overall I think this is close to ready. Adding the `captureAWS` functionality should be easy enough, just iterating over all clients and capturing them. 

Manual mode will be a bit trickier, I left a few ideas so let me know what you think. Please take a look when you're available!

packages/core/package.json
```diff
@@ -15,13 +15,18 @@
...
     "@types/cls-hooked": "*",
     "atomic-batcher": "^1.0.2",
     "cls-hooked": "^4.2.2",
     "pkginfo": "^0.4.0",
     "semver": "^5.3.0"
   },
   "devDependencies": {
+    "@aws-sdk/config-resolver": "^1.0.0-gamma.1",
+    "@aws-sdk/middleware-stack": "^1.0.0-gamma.1",
+    "@aws-sdk/types": "^1.0.0-gamma.1",
```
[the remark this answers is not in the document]
monken: We only import TypeScript types from these packages. They are only required for development.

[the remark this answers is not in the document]
monken: moved them to top-level package.json

monken:
@willarmiros `captureAWS` support is probably no longer necessary (or even possible). The v3 SDK recommends that you only import and install the clients you are actually using in your code.

packages/core/lib/patchers/aws3_p.ts
monken: I tried the middleware stack but was unable to capture any exceptions. The middleware stack only gave me access to the raw response so that would force us to build the exception in the patcher. I'd love for someone from the v3 team to chime in and propose a better solution.

packages/core/test/unit/patchers/aws3_p.test.js
monken: Good catch! It's definitely a problem because the presign packages add themselves as a middleware. I'll look into it!

packages/core/lib/patchers/aws3_p.ts
monken: `captureAWSClient` now takes a second, optional segment argument. That would address the `manual` mode. This means that for manual mode, you need to call `captureAWSClient` any time you create a new segment. That way we can ensure that we don't break the TypeScript experience for any developers.

packages/core/test/unit/patchers/aws3_p.test.js
willarmiros: I see. If we just add our middleware before theirs using the relative ordering would that fix it? If it's a middleware as opposed to a client can it get captured?

packages/core/lib/patchers/aws3_p.ts
willarmiros: Not sure if he'll respond but: @trivikr do you know of any better way to capture exceptions from a AWS SDK V3 request using the middleware stack? Otherwise we have to use a workaround by surrounding the requests with a `try/catch` like we are here.

packages/core/lib/patchers/aws3_p.ts
willarmiros: Hmmm this would be a pretty big break from the other types of manual mode instrumentation we have - calling `captureAWSClient` is generally a one-and-done deal. But then again it is called manual mode for a reason, I'll discuss this with the team as well.

packages/core/test/unit/patchers/aws3_p.test.js
monken: Unfortunately, the presign middleware inserts itself right before the `AwsAuthMiddleware` (https://github.com/aws/aws-sdk-js-v3/blob/master/packages/s3-request-presigner/src/getSignedUrl.ts#L46) so we cannot get in between anymore. The most straight forward implementation is to leave it up to the user. If they want to have the presigned URL instrumentalized (which I think is a feature not a bug) then they can add the XRay patcher.

```javascript
import { getSignedUrl } from "@aws-sdk/s3-request-presigner";
import { S3Client, GetObjectCommand } from "@aws-sdk/client-s3";
const client = new S3Client(clientParams); // feel free to wrap in captureAWSClient()
const command = new GetObjectCommand(getObjectParams);
const url = await getSignedUrl(client, command, { expiresIn: 3600 });
```

willarmiros: I'm a little confused, if we're instrumenting by replacing the `client.send()` function does `getSignedUrl` use that under the hood? Right now, we'll only be adding our middleware which adds a trace header which won't be actually used for anything. I don't think an actual subsegment will be made just for `getSignedUrl`.

monken: Yes, `getSignedUrl` calls `send` here https://github.com/aws/aws-sdk-js-v3/blob/master/packages/s3-request-presigner/src/getSignedUrl.ts#L51.
I also noticed that `getSignedUrl` removes headers from some other middlewares. I wonder if we could convince them to include the XRay header as well (https://github.com/aws/aws-sdk-js-v3/blob/master/packages/s3-request-presigner/src/getSignedUrl.ts#L26).
Like I said earlier, giving the developer the choice to instrumentalize a presigned request could be a feature rather than a bug. I'd make sure this behavior is well documented and let the developer choose. They can always create a new S3 client without the xray middleware applied when they presign URLs.

willarmiros: Ok thank you for pointing me to those links! I'd like to try this out and get a feel for it myself, but I realized that this API hasn't been exposed on your branch yet. Can you export it from our top-level `aws-xray` module and include an [automatic](https://github.com/aws/aws-xray-sdk-node/tree/master/packages/core#capture-outgoing-aws-requests-on-a-single-client) and [manual]() mode example in the readme like the ones I linked for V2? We already have the `captureAWSClient` API for v2 clients, so you'll need to change the function to something like `captureV3AWSClient`.

After I do some testing and we figure out the `getSignedUrl` experience we want, this should be good to go. Thanks for bearing with us!

willarmiros:
@monken agreed that it doesn't make sense to include `captureAWS`, at least not in this iteration. Do you think you'll be able to follow up on the other comments?

monken:
@willarmiros I added the documentation and exports. You should be able to start testing now.

packages/core/test/unit/patchers/aws3_p.test.js
entitycs: I have another quick question/consideration.

If all v3 clients implement packages/types/src/client.ts, is it necessary to have an outward-facing 'v3' version of the captureAWSClient method?  That is, could captureAWSClient check for the client interface implementation in the given argument, and forward the client to a "private" captureAWSv3Client method if the check passes?

(this message will also self-destruct)

willarmiros: After testing this I think capturing `getSignedUrl` as we are currently is the correct experience. I think I mainly misremembered previous customer issues that arose from presigining behavior thinking they wanted it removed, but that was not the case. Thanks for providing your input here!

willarmiros (commented):
Looking good @monken! A few nitpicks that arose from testing, but I tested using automatic and manual mode and for the most part things went smoothly. The only problem at this point is all the operation-specific attributes, but as mentioned that'll be out of scope for this PR.

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -0,0 +1,173 @@
...
+
+export function captureAWSClient<
+  Input extends object,
+  Output extends MetadataBearer,
+  Configuration extends DefaultConfiguration
+>(client: Client<Input, Output, Configuration>, manualSegment?: SegmentLike): Client<Input, Output, Configuration> {
+  // create local copy so that we can later call it
+  const send = client.send;
+
+  const serviceIdentifier = client.config.signingName;
```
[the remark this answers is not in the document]
workeitel: fyi I opened a feature request to the aws-sdk-js-v3 for an easier way to get service/operation names: https://github.com/aws/aws-sdk-js-v3/issues/2030

willarmiros:
@monken any updates on these comments?

packages/core/lib/patchers/aws3_p.ts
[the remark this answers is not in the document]
monken: we will need to require at least version 3.2.0 of the AWS V3 SDK. Otherwise we'll need to detect the version of the SDK and run dedicated code to detect the version. Also, the metadata object has changes with 3.2.0 so the changes would be quite significant to also support 3.1.0.

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -0,0 +1,173 @@
...
+  if (cause === 'fault') {
+    subsegment.addFaultFlag();
+  } else if (cause === 'error') {
+    subsegment.addErrorFlag();
+  }
+}
+
+type DefaultConfiguration = RegionInputConfig & {
+  signingName: string;
+};
```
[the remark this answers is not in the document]
monken: It's available with 3.2.0

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -0,0 +1,173 @@
...
+  client: any,
+  command: any,
+  metadata: ResponseMetadata,
+): Promise<[ServiceSegment, HttpResponse]> {
+  const { extendedRequestId, requestId, httpStatusCode: statusCode, attempts } = metadata;
+  const serviceIdentifier = client.config.signingName as string;
+
+  let operation: string = command.constructor.name.slice(0, -7);
+
+  const aws = new ServiceSegment(
```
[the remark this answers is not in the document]
monken: Good point. Maybe there is an opportunity to standardize the operation-specific attributes for all SDKs (not just JavaScript) and then use that as a template to generate the code.

packages/core/README.md
```diff
@@ -622,6 +641,24 @@ function sendRequest(host, cb, subsegment) {
 
 ### Capture outgoing AWS requests on a single client
 
+AWS SDK v3
+
+```js
+import { S3, PutObjectCommand } from '@aws-sdk/client-s3';
+
+// subsegment is an optional parameter that is required for manual mode
+// and can be omitted in automatic mode (e.g. inside a Lambda function).
```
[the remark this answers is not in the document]
monken: Added

monken:
Sorry for not being very responsive. I rebased and pushed a new version. Happy to squash the commits when you've reviewed them.

willarmiros:
@monken I believe you didn't re-compile the JS code to match the latest TS changes. Also need to update some tests for `serviceId`. Then we can merge this 🚀

wangzlei:
Kindly ask can I have aws sdk v3 support soon?

monken:
new PR #386

Issue #162, opened 2019-08-13
NodeJs HTTP request: limited captured data


jeanmatthieud:
Having the ability to capture HTTP(s) outgoing request is great, but the current implementation is quite limited for debugging.

It could be great to have at least HTTP headers (req / resp) included in the captured data.

awssandra:
Hi jeanmatthieud,

The concern here is that most of the headers would be considered non-actionable data, and would eat up a lot of segment data space. So it'd be not too useful to record all headers on every request automatically. But, I do think there is a larger story here on having a patch "configurer" to include extra data fields (sort of similar to the way we handle whitelisting AWS SDK fields for capturing) that could be configured to record not only header fields, but other fields of interest. Thoughts?

jeanmatthieud:
Thanks for your answer.
I like the idea of beeing able to choose which field to record. Like a function called for every request (with req/res as parameters) which will return additional data to record (as a map).

PR #299, opened 2020-05-29
Add callback to allow custom subsegment interaction when captureHTTPs is used

paulbrimicombe:
*Issue #162*

*Description of changes:*

* Add a callback to `captureHTTPs` and `captureHTTPsGlobal` that allows users to add metadata / annotations etc. to the subsegment based on the request, response and any error emitted.
* This seems like the most flexible way to allow additional information to be added to the subsegment.
* This feature is particularly useful when interacting with systems that are not integrated with XRay (e.g. want to record cache hit / miss headers, possibly on a per-host basis). Headers of this kind vary on a per-host basis which is why `req` is a parameter to the callback.

*Comments:*

* `captureHTTPs()` and `captureHTTPsGlobal()` appear to have an undocumented second parameter (at least it's not in the docs I've read) which makes the API for this new feature slightly less appealing.
* Having a single callback for the success and error case might be a little unwieldy. Would it be better to have two callbacks for the two cases or even to have some kind of event listener/emitter pattern?
* How defensive do we need to be about errors being thrown in the callback? Do we want to swallow/handle them? Or is it OK to let them propagate?
* The testing coverage is a little off — I've not covered the case where an error is emitted and the callback isn't provided just because that was a more intrusive change to the existing tests.
* Is there any more documentation that I should be updating.

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -460,6 +460,28 @@ This creates 5 nested subsegments on the root segment and captures timing data i
     // those used by third party modules, will now be traced
     var http = require('http');
 
+### Capture all outgoing HTTP and HTTPS requests, adding custom subsegment information
+
+    const callback = (subsegment, req, res, err) => {
+      subsegment.addMetadata('accept', req.getHeader('accept'));
+
+      if (err && err.code) {
+        subsegment.addAnnotation('errorCode', err.code);
+      }
+
+      if (res) {
+        subsegment.addMetadata('content-type', res.getHeader('content-type'));
+      }
+    };
+    AWSXRay.captureHTTPsGlobal(require('http'), null, callback);
+    AWSXRay.captureHTTPsGlobal(require('https'), null, callback);
+
+    // Requests with this http client, and any other http/https client including
+    // those used by third party modules, will now be traced
+    // Additional metadata / annotations can be added in the callback based on 
+    // the request, response and any error
+    var http = require('http');
+
 ### Capture outgoing HTTP/S requests with a traced client
 
     //returns a copy of the http module that is patched, can patch https as well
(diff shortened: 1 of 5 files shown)


willarmiros:
Hi @paulbrimicombe, 
Thank you very much for your contribution! I'll be taking a look at it soon. As for your comments:
* Yeah, that extra parameter is only documented in the [API reference](https://docs.aws.amazon.com/xray-sdk-for-nodejs/latest/reference/module-http_p.html), and is very seldom used by customers. But it does have a use (explained [here](https://forums.aws.amazon.com/thread.jspa?threadID=312851&tstart=25)) so we'll have to leave API as is I think.
* I think a single callback is acceptable, since adding an additional one would make the `captureHTTPs` API even more unwieldy. A listener/emitter pattern could be problematic since we want customers to have to write as little code as possible for the X-Ray SDK, and introducing listeners specific to the SDK inside their HTTP response handlers might be a little too much to ask.
* IMO the SDK shouldn't have to be responsible for handling errors in user-provided code. So I think it's ok as is.
* I'll comment on test coverage in the review
* The API reference will be automatically updated on the next release with your addition to the JSDOC, so should be good on the docs front.

paulbrimicombe:
@willarmiros many thanks for the response. Do you think you'll be able to review the changes some time soon please?

paulbrimicombe:
That's great. Thanks @willarmiros !

paulbrimicombe:
Any idea when this will get released @willarmiros ?

willarmiros:
Unfortunately we can't share any official timelines, however this will be included in the next release (3.1.0) so feel free to watch the repo to be notified of it!

PR #313, opened 2020-06-23
guard against undefined req.agent

### LINKED ISSUE #312: guard against undefined req.agent
We're seeing a crash because `req.agent` is undefined at [remote_request_data.js#L17](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/segments/attributes/remote_request_data.js#L17
). The underlying problem could certainly be ours but regardless I think some defensive coding would be advised:

```
url: (req.agent.protocol + '//' + req.getHeader('host') +  stripQueryStringFromPath(req.path)) || '',
```

joebowbeer:
*Issue #, if available:*

Fixes #312 

*Description of changes:*

Guards against undefined req.agent

### DIFF
--- a/packages/core/lib/segments/attributes/remote_request_data.js
+++ b/packages/core/lib/segments/attributes/remote_request_data.js
@@ -14,7 +14,7 @@ function RemoteRequestData(req, res, downstreamXRayEnabled) {
 
 RemoteRequestData.prototype.init = function init(req, res, downstreamXRayEnabled) {
   this.request = {
-    url: (req.agent.protocol + '//' + req.getHeader('host') +  stripQueryStringFromPath(req.path)) || '',
+    url: ((req.agent && req.agent.protocol) + '//' + req.getHeader('host') +  stripQueryStringFromPath(req.path)) || '',
     method: req.method || '',
   };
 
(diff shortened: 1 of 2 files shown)


packages/core/lib/segments/attributes/remote_request_data.js
```diff
@@ -14,7 +14,7 @@ function RemoteRequestData(req, res, downstreamXRayEnabled) {
 
 RemoteRequestData.prototype.init = function init(req, res, downstreamXRayEnabled) {
   this.request = {
-    url: (req.agent.protocol + '//' + req.getHeader('host') +  stripQueryStringFromPath(req.path)) || '',
+    url: ((req.agent && req.agent.protocol) + '//' + req.getHeader('host') +  stripQueryStringFromPath(req.path)) || '',
```
willarmiros, line 17: While this avoids throwing the error, in the event that `req.agent` is undefined it would print the URL as:

```
undefined//example.com/
```

I'd advise changing this to: 

```js
url: (req.agent && req.agent.protocol) ? (req.agent.protocol + '//' + req.getHeader('host') +  stripQueryStringFromPath(req.path)) : '',	
```

joebowbeer: Done; looking at the original line again, I see that it contained an undefined check that would never be executed.

joebowbeer:
PTAL

Issue #318, opened 2020-07-06
http client error events might be suppressed


Flarna:
I think [this check](https://github.com/aws/aws-xray-sdk-node/blob/0590805fe2591ba7cc2c6891caec4ef43fed40e1/packages/core/lib/patchers/http_p.js#L141) in http_p.js should be improved as it may result in not re-emitting an error event.

`this._events.error` is either an Array (if more then 1 listener is installed) or `Function` (only 1 listener installed).
Currently this check works because the installed listener `errorCapturer` has exactly 1 named parameter therefore `errorCapturer.length === 1`.
But if there is a module used which wraps the installed listeners (e.g. for CLS,...) or if `once` is used to install the listener this is broken. Besides that it relies on internals of `EventEmitter` (the `_events` field`).

I recommend to use the public `EventEmitter` Api to check that <= 1 listener is installed.

Recent NodeJs versions allow to listen on [errorMonitor](https://nodejs.org/dist/latest-v12.x/docs/api/events.html#events_eventemitter_errormonitor) to avoid the need to re-emit in case only 1 listener is present at all.

willarmiros:
Hi @Flarna, 
Thank you for pointing out this potential vulnerability. It appears that `errorMonitor` is only available in Node 12 and later, but this SDK still supports Node 10. So do you think a valid alternative would be to use the [listenerCount](https://nodejs.org/dist/latest-v12.x/docs/api/events.html#events_emitter_listenercount_eventname) API, e.g. 

```js
if (this.listenerCount('error') === 1) {
  this.removeListener('error', errorCapturer);
  this.emit('error', e);
}
```

Flarna:
What about the following to use `errorMonitor` if present and fall back to `error` and reemit if not, e.g.:
```
req.on(req.errorMonitor || 'errror', errorCapturer)
```
and in listener:
```
if ((this.errorMonitor != null) && (this.listenerCount('error') <= 1)) {
  this.removeListener('error', errorCapturer);
  this.emit('error', e);
}
```
Main issue is that there are other modules out in the wild using the same "if I'm the only listener remove myself and reemit" algorithm which works fine as long as you are alone.

Reason for `listenerCount() <= 1` instead `=== 1` is that I got hit by some module which also patched EventEmitter and there was some race between my and there code ending up in `listenerCount` reporting 0 even I was in in the middle of my listener :o).

willarmiros:
[...]

Issue #18, opened 2018-01-26
HTTP Patcher leaks response & stalls segments if no other response listener is supplied


omsmith:
Ran into this while testing the examples I was providing in #11. Not sure how much real-world impact this has, but the http patcher adds a _response_ listener to outgoing http requests, and then sets up a listener for `res.on('end', ...)` in order to close the subsegment. The problem is, once you do add a _response_ listener, then the response _must_ be consumed.

From the [http documentation](https://nodejs.org/api/http.html#http_class_http_clientrequest):
> If no 'response' handler is added, then the response will be entirely discarded. However, if a 'response' event handler is added, then the data from the response object must be consumed, either by calling response.read() whenever there is a 'readable' event, or by adding a 'data' handler, or by calling the .resume() method. Until the data is consumed, the 'end' event will not fire. Also, until the data is read it will consume memory that can eventually lead to a 'process out of memory' error.

Currently if the developer didn't add a listener of their own, the response is never consumed, meaning the 
response data is leaked, _and_ since `end` is never fired, the subsegment will never close meaning it and its parent (sub)segment will never be flushed and sent to the daemon.

Reason I'm not sure how much real-world impact this would have, is I'm not sure how often you would simply not care about the response, but I'm sure there's some sort of fire-and-forget systems out there...?

The following _should_ fix it, but I've not developed any tests for it. It consumes the body if a callback wasn't provided, and if x-ray is the only response listener. Meaning neither of the two scenarios below occurred:

```js
http.request('http://example.com', res => res.resume()).end();
http.request('http://example.com').on('response', res => res.resume()).end();
```

```diff
diff --git a/packages/core/lib/patchers/http_p.js b/packages/core/lib/patchers/http_p.js
index 5c0cb68..f5311ef 100644
--- a/packages/core/lib/patchers/http_p.js
+++ b/packages/core/lib/patchers/http_p.js
@@ -137,6 +137,8 @@ function enableCapture(module, downstreamXRayEnabled) {
         } else {
           callback(res);
         }
+      } else if (req.listenerCount('response') === 1) {
+        res.resume();
       }
     }).on('error', errorCapturer);
 ```

Of course, if the developer _does_ add their own listener and neglects to consume the response, then you still run into the segment never closing, but that's a different issue, with developer error mixed in.

haotianw465:
[...]

PR #321, opened 2020-07-09
replaced error listeners with errorMonitor

willarmiros:
*Issue #, if available:*
#318, #18 

*Description of changes:*
Changed our `errorCapturer` to listen on the `errorMonitor` event instead of the `error` event. As pointed out in related issue, the old way could be unreliably re-emitting errors to customer code. Since `errorMonitor` was only introduced in Node 12, we will have to fall back to the old way of error capturing in Node 10 until it is deprecated.

Also added a unit test and tested on my sample app to verify behavior.

*EDIT*
Also added other fix described in comment below.

### DIFF
--- a/packages/core/lib/patchers/http_p.js
+++ b/packages/core/lib/patchers/http_p.js
@@ -161,12 +161,16 @@ function enableCapture(module, downstreamXRayEnabled, subsegmentCallback) {
 
         var cause = Utils.getCauseTypeFromHttpStatus(res.statusCode);
 
+        console.log('closing!!');
+
         if (cause)
           subsegment[cause] = true;
 
         subsegment.addRemoteRequestData(res.req, res, !!downstreamXRayEnabled);
         subsegment.close();
       });
+
+      // console.log(res.req.listenerCount('response'));
 
       if (typeof callback === 'function') {
         if (contextUtils.isAutomaticMode()) {
@@ -179,10 +183,11 @@ function enableCapture(module, downstreamXRayEnabled, subsegmentCallback) {
         } else {
           callback(res);
         }
-        // if no callback provided by user application, then we consume the response
-        // so the 'end' event fires
+        // if no callback provided by user application, AND no explicit response listener
+        // added by user application, then we consume the response so the 'end' event fires
         // See: https://nodejs.org/api/http.html#http_class_http_clientrequest
-      } else {
+      } else if (res.req && res.req.listenerCount('response') === 0) {
+        console.log('resuming!');
         res.resume();
       }
     });
--- a/packages/core/lib/patchers/http_p.js
+++ b/packages/core/lib/patchers/http_p.js
@@ -142,7 +142,7 @@ function enableCapture(module, downstreamXRayEnabled, subsegmentCallback) {
       // Only need to remove our listener & re-emit if we're not listening using the errorMonitor,
       // e.g. the app is running on Node 10. Otherwise the errorMonitor will re-emit automatically.
       // See: https://github.com/aws/aws-xray-sdk-node/issues/318
-      // TODO: Remove this logic when the 'error' listener is removed once node 10 support is deprecated
+      // TODO: Remove this logic once node 12 support is deprecated
       if (!events.errorMonitor && this.listenerCount('error') <= 1) {
         this.removeListener('error', errorCapturer);
         this.emit('error', e);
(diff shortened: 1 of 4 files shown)


willarmiros:
@Flarna feel free to give this a look :)

Flarna:
Implementation looks correct.
One note regarding presence of `events.errorMonitor` - it was introduced with 12.17.0 so it's not enough to wait until 10.x is EOL.

willarmiros:
@Flarna changed some wording w.r.t. Node 12.17. 

Also, for #18, I kept getting 0 for the `listenerCount('response')` because I was doing it from within the callback of the `request` function. I assume that's because the callbacks had not yet been registered within the response callback. However, I would get a `listenerCount('response')` of 1 when I explicitly added a `req.on('response')` listener. 

So, to cover the 2 cases you [commented](https://github.com/aws/aws-xray-sdk-node/issues/18#issuecomment-654995261), we will only consume the response on behalf of the user if both:
1. They did not provide a callback, as checked by `typeof callback === 'function'`
2. They did not add an explicit `response` listener to their request object, as checked by the `listenerCount === 0`

Flarna:
> Also, for #18, I kept getting 0 for the `listenerCount('response')` because I was doing it from within the callback of the `request` function. I assume that's because the callbacks had not yet been registered within the response callback. However, I would get a `listenerCount('response')` of 1 when I explicitly added a `req.on('response')` listener.

If you pass a callback to `http.get` it's registered [as once listener](https://github.com/nodejs/node/blob/v12.18.2/lib/_http_client.js#L194) therefore in the callback it's already removed as EventEmitter [first removes the listener and then calls it](https://github.com/nodejs/node/blob/v12.18.2/lib/events.js#L416-L424).

There are following cases to my understanding:
* `http.get(<url>, cb)` => cb will be registered via `once`, you will see 0 listeners and _should not_ dump data
* `http.get(<url>)` => your injected callback will be installed via `once`, you will see 0 listeners and _should_ dump the data
* `http.get(<url>).on('response', cb)` => cb will be registered via `on`, you will see 1 listener and _should not_ dump data
* `http.get(<url>).once('response', cb)` => cb will be registered via `once`, you will see 0 listener and _should not_ dump data

willarmiros:
Ahh it's added as a `once` listener that makes sense. I agree with your cases, I believe they are all handled. In the order that you laid them out, here's how the X-Ray SDK handles them:

* [callback](https://github.com/aws/aws-xray-sdk-node/pull/321/files#diff-5f8797bd0d2745c5647f4c0e258805baR171) passed to us by user code is defined, so we *do not* dump data
* callback passed to us is undefined, and response listeners [will be zero](https://github.com/aws/aws-xray-sdk-node/pull/321/files#diff-5f8797bd0d2745c5647f4c0e258805baR185), so we *do* dump data
* callback passed to us is undefined but response listeners will be 1, so we *do not* dump data
* callback passed to us is undefined but response listeners will be 1, so we *do not* dump data

On the last case, I checked with the below code and verified that the X-Ray SDK's callback *did* actually register the `once('response', cb)` listener when it was added by the user. I think that that's because our callback is executed before the user's, so their once listener has not yet been removed.

```js
https.get('https://google.com')
  .once('response', httpRes => {
    console.log('got response');
    // consume response, because X-Ray SDK doesn't
  });
```

packages/core/lib/patchers/http_p.js
```diff
@@ -138,7 +139,11 @@ function enableCapture(module, downstreamXRayEnabled, subsegmentCallback) {
         subsegment.close(e);
       }
 
-      if (this._events && this._events.error && this._events.error.length === 1) {
+      // Only need to remove our listener & re-emit if we're not listening using the errorMonitor,
+      // e.g. the app is running on Node 10. Otherwise the errorMonitor will re-emit automatically.
+      // See: https://github.com/aws/aws-xray-sdk-node/issues/318
+      // TODO: Remove this logic once node 12 support is deprecated
```
shengxil, line 145: Do you mean "once 10 support is deprecated"?

shengxil: Do you mind explaining more on why do we remove the error listener  when `listenerCount('error')` returns 0?

packages/core/lib/patchers/http_p.js
```diff
@@ -138,7 +139,11 @@ function enableCapture(module, downstreamXRayEnabled, subsegmentCallback) {
...
+      if (!events.errorMonitor && this.listenerCount('error') <= 1) {
         this.removeListener('error', errorCapturer);
```
shengxil, line 147: What happens when `this.listenerCount('error')` return 0 (which semantically means there is no error listener) and we call `this.removeListener('error', errorCapturer);`? Would there be an error?

packages/core/lib/patchers/http_p.js
```diff
@@ -138,7 +139,11 @@ function enableCapture(module, downstreamXRayEnabled, subsegmentCallback) {
         subsegment.close(e);
       }
 
-      if (this._events && this._events.error && this._events.error.length === 1) {
+      // Only need to remove our listener & re-emit if we're not listening using the errorMonitor,
+      // e.g. the app is running on Node 10. Otherwise the errorMonitor will re-emit automatically.
+      // See: https://github.com/aws/aws-xray-sdk-node/issues/318
+      // TODO: Remove this logic once node 12 support is deprecated
+      if (!events.errorMonitor && this.listenerCount('error') <= 1) {
```
shengxil, line 146: What happens when this.listenerCount('error') return 0 (which semantically means there is no error listener) and we call this.removeListener('error', errorCapturer);? Would there be an error?

willarmiros: Nope, according to the node docs `removeListener` removes *at most* one listener, so it just no-ops if there's none to remove.

https://nodejs.org/dist/latest-v12.x/docs/api/events.html#events_emitter_removelistener_eventname_listener

PR #328, opened 2020-08-27
[aws-xray-sdk-mysql] Handle operations that return a promise instead of an event emitter (mysql2)

TuckerWhitehouse:
*Issue #, if available:*
https://github.com/aws/aws-xray-sdk-node/issues?q=mysql2

*Description of changes:*
The `mysql2` package returns a promise instead of a wrapped object with an event emitter. Fortunately, the two events this package relies on (`end`, `error`) can be mapped to the promise being resolved or rejected.

### DIFF
--- a/packages/mysql/lib/mysql_p.js
+++ b/packages/mysql/lib/mysql_p.js
@@ -90,7 +90,7 @@ function patchOf(poolCluster) {
 
     var resultPool = poolCluster[baseFcn].apply(poolCluster, args);
     return patchObject(resultPool);
-  }
+  };
 }
 
 function patchGetConnection(pool) {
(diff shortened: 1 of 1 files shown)


willarmiros (approved):
Awesome improvement @TuckerWhitehouse! I know that workarounds existed to support mysql2 with our existing mysql patcher, but after this change are there any further changes you think are required for complete support?

TuckerWhitehouse:
Hey @willarmiros - thanks! I've branch deployed with this code a couple of times just to validate things and can confirm I'm seeing xray traces for sql queries as expected. The code I'm testing with only uses `.query`, but I think this change would work for `.execute` all the same :)

Do you know how often new releases are cut or when a release with this change would be available?

willarmiros:
Awesome to hear! We can't give exact dates for releases but I will prioritize making a release with this feature in the near future.

PR #336, opened 2020-09-12
Handle requests without agent to fallback to request.protocol

khuezy:
*Issue #, if available:*
Relates to https://github.com/aws/aws-xray-sdk-node/issues/312

*Description of changes:*
Hi, I've encountered a crash when a request didn't have an agent on 3.1.0. I see that it's been fixed in 3.2.0, but this PR attempts to fallback to the `request.protocol` instead of returning an empty string if the agent isn't available.

### DIFF
--- a/packages/core/lib/segments/attributes/remote_request_data.js
+++ b/packages/core/lib/segments/attributes/remote_request_data.js
@@ -13,8 +13,11 @@ function RemoteRequestData(req, res, downstreamXRayEnabled) {
 }
 
 RemoteRequestData.prototype.init = function init(req, res, downstreamXRayEnabled) {
+  // Not all requests have "agent"
+  const protocol = req.agent ? req.agent.protocol: req.protocol + ':'
+  
   this.request = {
-    url: (req.agent && req.agent.protocol) ? (req.agent.protocol + '//' + req.getHeader('host') +  stripQueryStringFromPath(req.path)) : '',
+    url: protocol + '//' + req.getHeader('host') +  stripQueryStringFromPath(req.path),
     method: req.method || '',
   };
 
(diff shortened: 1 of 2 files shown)


packages/core/lib/segments/attributes/remote_request_data.js
```diff
@@ -13,8 +13,11 @@ function RemoteRequestData(req, res, downstreamXRayEnabled) {
 }
 
 RemoteRequestData.prototype.init = function init(req, res, downstreamXRayEnabled) {
+  // Not all requests have "agent"
+  const protocol = req.agent ? req.agent.protocol: req.protocol + ':'
```
willarmiros, line 17: Nitpick: Can you add a space after `req.agent.protocol`

willarmiros, line 17: Is it guaranteed that if `req.agent` is absent `req.protocol` will be defined, or should we add an additional check?

Also, is it documented anywhere that `req.protocol` doesn't have a trailing colon but `req.agent.protocol` does? Seems strange that they're different in that way.

willarmiros (commented):
Thanks so much for the contribution @khuezy! Left a couple comments for you.

packages/core/lib/segments/attributes/remote_request_data.js
khuezy: Good point, `req.protocol` was added in `14.5.0`, so we'll need additional checks.
I haven't seen any documentation, but the source code has the agent protocol defaulting to `https:` https://github.com/nodejs/node/blob/8e173836538efb95f5ac5e898f177efff681d2aa/lib/https.js#L154

willarmiros:
Hi @khuezy do you still plan on finishing this PR up?

khuezy:
Hi sorry, was busy w/ life. I'll close and reopen when I get some time.

Issue #335, opened 2020-09-11
PG capture does not work well with queryConfig objects.


nmaves:
Per the PG documentation, you should be able to pass a queryConfig object with the shape as the following.  This is not working with the AWSXRay.capturePostgres().  I think the issue is around the fact that we are not passing a callback in but that should trigger the promise returned and it is not.

```
interface QueryConfig {
  // the raw query text
  text: string;
  // an array of query parameters
  values?: Array<mixed>;
  // name of the query - used for prepared statements
  name?: string;
  // by default rows come out as a key/value pair for each row
  // pass the string 'array' here to receive rows as an array of values
  rowMode?: string;
  // custom type parsers just for this query result
  types?: Types;
}
```

nmaves:
Looks like you can just update this https://github.com/aws/aws-xray-sdk-node/blob/bd2a8f2554c3969a2b47afb59c1a7085b54f3cd8/packages/postgres/lib/postgres_p.js#L41

to be the following

```
args.callback = argsObj[2] || argsObj[1] || argsObj[0].callback;
```

awssandra:
Hi nmaves,

Thanks so much for the deep dive! If you'd like, you can open a PR and we can merge this in, and we can credit you as a contributor.

Can you confirm which version of PG you're using? We'll need to make sure we keep backwards compatibility.

Thanks!
Sandra

nmaves:
@awssandra We are using node-postgres 8.3.3

PR #337, opened 2020-09-16
Promise callback with queryConfig options.

nmaves:
Resolves https://github.com/aws/aws-xray-sdk-node/issues/335

When you execute a PG query with an QueryConfig object the implicit promise callback is the 3 element in the args.

### DIFF
--- a/packages/postgres/lib/postgres_p.js
+++ b/packages/postgres/lib/postgres_p.js
@@ -38,7 +38,7 @@ function resolveArguments(argsObj) {
     if (argsObj[0] instanceof Object) {
       args.sql = argsObj[0].text;
       args.values = argsObj[0].values;
-      args.callback = argsObj[1] || argsObj[0].callback;
+      args.callback = typeof argsObj[1] === 'function' ? argsObj[1] : (typeof argsObj[2] === 'function' ? argsObj[2] :  argsObj[0].callback);
     } else {
       args.sql = argsObj[0];
       args.values = typeof argsObj[1] !== 'function' ? argsObj[1] : null;


willarmiros (commented):
Thanks for the contribution @nmaves! Any chance you could add a unit test to verify this behavior when capturing a query with a promise-based callback? Also, what would the 2nd argument be then in this case?

nmaves:
Honestly, I am not sure how to write this with your current test setup.

From my testing that second argument is null.  I put in debug statements with your code and found that to be the case.

willarmiros:
That's fair enough, I'd be fine with not having a unit test here if there was some validation to check if the 3rd argument is a function. Something similar to what we already do on L45: https://github.com/aws/aws-xray-sdk-node/blob/master/packages/postgres/lib/postgres_p.js#L45

nmaves:
Does that work?

willarmiros:
Looks like the test is failing because you need to include `argsObj[0].callback` instead of `undefined` as the final fallback possibility.

nmaves:
The test is now passing.

willarmiros:
Great, thanks again for the changes!

willarmiros (approved):

nmaves:
hey @willarmiros you guys think you might snap a release anytime soon to get this out in the wild?

willarmiros:
We're definitely due for a release, I'll prioritize one soon. Thanks for the ping @nmaves !

nmaves:
@willarmiros don't make me beg and look like a need open source abuser :)

PR #340, opened 2020-10-08
Fix: allow all mysql query options

haversnail-cbs:
*Issue #, if available:* 
**N/A** (figured I'd just submit the PR vs. open another issue) 😉 

*Description of changes:*
Updates the `mysql` package to support all query options by passing along the entire object to the original call. This fixes a bug where any additional options provided when calling `query` (e.g. `typeCast`, `nestTables`, etc.) are getting ignored.

### DIFF
--- a/packages/mysql/lib/mysql_p.js
+++ b/packages/mysql/lib/mysql_p.js
@@ -141,12 +141,7 @@ function resolveArguments(argsObj) {
 
   if (argsObj && argsObj.length > 0) {
     if (argsObj[0] instanceof Object) {
-      args.sql = argsObj[0].sql;
-
-      if (argsObj[0].timeout) {
-        args.timeout = argsObj[0].timeout;
-      }
-
+      args.sql = argsObj[0];
       args.values = argsObj[0].values;
       args.callback = argsObj[1];
       if (!argsObj[1] && argsObj[0].on instanceof Function) args.sql = argsObj[0];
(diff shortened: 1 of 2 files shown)


packages/mysql/lib/mysql_p.js
```diff
@@ -141,12 +141,7 @@ function resolveArguments(argsObj) {
 
   if (argsObj && argsObj.length > 0) {
     if (argsObj[0] instanceof Object) {
-      args.sql = argsObj[0].sql;
-
-      if (argsObj[0].timeout) {
-        args.timeout = argsObj[0].timeout;
-      }
-
+      args.sql = argsObj[0];
```
willarmiros, line 144: Doesn't this set `args.sql` to an entire argument object rather than just the query string as it was before? I don't have any problem with this, I'm just confused by how the same behavior is maintained. Specifically want to make sure that calling the original operation will still work if someone is using `mysql` in the mode captured by the `else` block, where they pass the query string as the first parameter rather than an arguments object.

willarmiros (commented):
Thanks so much for the contribution! This is a great generalization of the PR we had to include the timeout parameter a while back. Just want to do a sanity check.

packages/mysql/lib/mysql_p.js
haversnail-cbs: @willarmiros that's correct, since the [original `query` function](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mysql/index.d.ts#L384) accepts an object as its first argument, this just ensures that the entire object does in fact get passed along (vs. just the `sql` and `timeout` properties). That's a fair concern though; this condition could definitely be made simpler and easier to reason about now that we're passing along the entire object, however I decided to leave things as-is for now so as not to overcomplicate this PR. The logic in the `else` block seems like it should support the query string flavor just fine though, and the tests all passed w/o issue. 👍 Either way, let me know if this needs any clarification; happy to discuss further.

willarmiros (approved):
Thanks for the explanation!

PR #345, opened 2020-10-21
feat:IGNORE_ERROR context missing strategy

cheruvian:
Adds a new `IGNORE_ERROR` context missing strategy, this is especially useful when running Lambda code/libraries locally to not pollute logs with many lines of context missing.

### DIFF
--- a/packages/core/test/unit/context_utils.test.js
+++ b/packages/core/test/unit/context_utils.test.js
@@ -9,6 +9,8 @@ var LOG_ERROR = 'LOG_ERROR';
 var LOG_ERROR_FCN_NAME = 'contextMissingLogError';
 var RUNTIME_ERROR = 'RUNTIME_ERROR';
 var RUNTIME_ERROR_FCN_NAME = 'contextMissingRuntimeError';
+var IGNORE_ERROR = 'IGNORE_ERROR';
+var IGNORE_ERROR_FCN_NAME = 'contextMissingIgnoreError';
 
 describe('ContextUtils', function() {
   function reloadContextUtils() {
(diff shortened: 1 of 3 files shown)


packages/core/test/unit/context_utils.test.js
```diff
@@ -146,7 +148,12 @@ describe('ContextUtils', function() {
 
     it('should accept and set the RUNTIME_ERROR strategy', function() {
       ContextUtils.setContextMissingStrategy(RUNTIME_ERROR);
-      assert.notEqual(ContextUtils.contextMissingStrategy.contextMissing, RUNTIME_ERROR_FCN_NAME);
+      assert.equal(ContextUtils.contextMissingStrategy.contextMissing.name, RUNTIME_ERROR_FCN_NAME);
```
willarmiros, line 151: Not sure what was going on in this test before, thanks :)

willarmiros (approved):
Thanks for the addition!

willarmiros:
The Travis failure is a problem in the Koa patcher and only occurs in Node 15, and it is unrelated to this PR. We should fix the issue in the Koa ASAP to not have these misleading CI failures.

isaacl:
Nice.  For anyone who wants this behavior early, here is a workaround:

- Set AWS_XRAY_CONTEXT_MISSING to empty string.  This is only possible programmatically.  For example, with serverless yaml: `AWS_XRAY_CONTEXT_MISSING: ''`
- Run `xray.setContextMissingStrategy(() => undefined)` in your code.  This is a NO-OP in lambda unless you do the environment hack above.

willarmiros:
Thanks for the pointer! The only thing I'd add is with respect to:

> Set AWS_XRAY_CONTEXT_MISSING to empty string. This is only possible programmatically

I've verified you can also set the environment variable to no value in the Lambda Console for the same effect, if you're not using any kind of infra-as-code.

cheruvian:
> Nice. For anyone who wants this behavior early, here is a workaround:
> 
> * Set AWS_XRAY_CONTEXT_MISSING to empty string.  This is only possible programmatically.  For example, with serverless yaml: `AWS_XRAY_CONTEXT_MISSING: ''`
> * Run `xray.setContextMissingStrategy(() => undefined)` in your code.  This is a NO-OP in lambda unless you do the environment hack above.

You can always programmatically set a custom context missing strategy if you need this early

```
setContextMissingStrategy(() => undefined)
```

PR #350, opened 2020-10-24
types(segment): add end_time to Segment

RafaelGSS:
*Description of changes:*

I use `xray` creating the wrapper manually, and normally have something like:

```ts
const segment = new AWSXRay.Segment(rootName);
const ns = AWSXRay.getNamespace();
await ns.runPromise(async() => {
  segment.addIncomingRequestData(this.createFakeIncomingRequest(rootName));
  AWSXRay.setSegment(segment);
  await func(); // Original function
  segment.close();
  console.log({
    traceId: segment.trace_id,
    elapsed: +(segment as any).end_time - +segment.start_time, // This PR solve the (segment as any) as workaround
  });
```

To prevents the `segment as any` I mapped the `end_time` on Segment class.

### DIFF
--- a/packages/core/lib/segments/segment.d.ts
+++ b/packages/core/lib/segments/segment.d.ts
@@ -5,6 +5,7 @@ declare class Segment {
   id: string;
   name: string;
   start_time: number;
+  end_time?: number;
   in_progress?: boolean;
   trace_id: string;
   parent_id?: string;


willarmiros (approved):
Thanks for the contribution @RafaelGSS! The CI failure is unrelated, but I'm going to try to fix it before merging.

willarmiros:
Given the small surface area of the change and since #351 might take some time to merge, I will merge this for now.

PR #351, opened 2020-10-26
updated dev dependencies and moved them to top-level

willarmiros:
*Description of changes:*
This PR attempts to fix the unit test that has been causing builds to fail recently ([example](https://travis-ci.org/github/aws/aws-xray-sdk-node/jobs/738608592)). The root cause of the failures appeared to be that `npx lerna bootstrap --hoist` was failing because of different versions among packages with duplicate `devDependencies`. This caused different versions of dev dependencies to be resolved in Travis than the ones resolved locally. 

To fix this, I used [lerna link convert](https://github.com/lerna/lerna#common-devdependencies) to pull all `devDependencies` to the top level and update them to the latest version. Now we have all our dev dependencies listed a single time, at a single version. Now `lerna bootstrap --hoist` works and we get all the benefits listed in that link. Only `devDependencies` and not regular `dependencies` were moved, so it will not impact end consumers since they don't install/need them anyway, and similarly it will not impact the size of the bundle since `devDependencies` are not vended in the final distribution either.

This is a mixed blessing, because it means all our packages share the same `devDependencies`. This is nice because it means we only have to list them once and there's no possibility of dependency hell from different packages depending on different versions of the same `devDependency`, and if we want to upgrade any of them it's a single line change. The downside is it is much harder to tell now which package(s) use which `devDependencies`. But I can't think of any situations where that *really* matters and where "Find All Usages" in an IDE wouldn't suffice.

Updating the `devDependencies` to latest required adapting our tests to some breaking changes, most of which were just changing `sinon.sandbox.create()` -> `sinon.createSandbox()`. So the JavaScript changes in this PR can be skimmed over.


packages/test_express/package.json
```diff
@@ -14,18 +14,8 @@
...
-    "aws-xray-sdk-express": "3.2.0",
-    "chai": "^3.5.0",
-    "cls-hooked": "^4.2.2",
-    "eslint": "^4.18.2",
-    "express": "^4.16.4",
-    "mocha": "^3.5.3"
-  },
   "scripts": {
-    "test": "mocha --recursive ./test/ -R spec"
+    "test": "mocha --recursive --exit ./test/ -R spec"
```
srprash, line 18: Not sure why is the `--exit` flag used here and not in other package?

willarmiros: Great question. This package has tests that spin up localhost servers (like WireMock), and even though the servers are closed at the end of all the unit tests (at least as far as I can tell) they still caused the test execution to hang. The `--exit` flag forces the test execution to exit once all the unit tests are complete.

The reason this was only a problem in this PR is because `mocha`, the testing framework, had this `--exit` flag as the default behavior in the version we were using before, then when we upgraded in this PR and the behavior became opt in. We can include the `--exit` flag in all the packages for consistency, what do you think?

srprash: Ah I see! Well, according to the mocha [doc](https://mochajs.org/#-exit) the use of `--exit` flag is required only if the tests hang and is not really recommended. If the tests in other packages do not hang, I don't see a necessity to add the `--exit` flag.

srprash (approved):
Looks good. :)

Issue #352, opened 2020-10-26
Migrate to GitHub Actions for CI


willarmiros:
We are upgrading many of our repos to GitHub Actions since Travis CI has been buggy and slow lately, and GH actions is more extensible (e.g. for automated releases). We should do the same here.

willarmiros:
Closing now that migration is complete and GH actions is working for normal PRs (e.g. #339).

PR #353, opened 2020-10-27
Replaced travis with github actions

willarmiros:
*Issue #, if available:*
#352

*Description of changes:*
Got very fed up with waiting 3+ hours for Travis builds to start so this PR replaces Travis with the GitHub actions. We'll run tests on the same versions of Node but on the 3 major OS's, like our other repos.


willarmiros:
Just realized that GH actions won't actually trigger for this PR until it's merged into the repo

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,37 @@
+name: Node.js SDK Continuous Build
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
```
NathanielRN, line 7: Should run tests on all branches so that they will run on PRs as well?

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,37 @@
...
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
+        node-version:
+          - 10.x
+          - 12.x
+          - 14.x
+          - 15.x
+    steps:
+      - name: Checkout Repository
```
NathanielRN, line 25: I think it makes sense to be verbose here for people starting with Github actions
```suggestion
      - name: Checkout AWS XRay SDK Node Repository @ default branch latest
```

NathanielRN (approved):
Thanks for this change!

.github/workflows/pr-build.yml
willarmiros: We don't have any branches besides `master`. This will trigger the workflow for all PRs made *against* the `aws:master` branch, regardless of their source branch. (It's not running on this PR because the workflow isn't merged in yet, but will on all future PRs)

anuraaga: Yup - I recommend this since otherwise the build runs on both the PR and the push to the fork - lots of waste of GitHub resources :)

.travis.yml
```diff
@@ -1,21 +0,0 @@
-language: node_js
```
anuraaga, line 1: 👋

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,37 @@
...
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
+        node-version:
+          - 10.x
+          - 12.x
+          - 14.x
+          - 15.x
+    steps:
+      - name: Checkout AWS XRay SDK Node Repository @ default branch latest
```
anuraaga, line 25: Can you add actions/cache? https://github.com/actions/cache/blob/main/examples.md#node---npm

anuraaga: Or maybe this one, not sure why there are two sections for npm and lerna

https://github.com/actions/cache/blob/main/examples.md#node---lerna

willarmiros: So I was hesitant to add caching because that was actually one fo the causes of failures in the flaky Travis tests: [this build](https://travis-ci.org/github/aws/aws-xray-sdk-node/builds/739341341/config) failed while [this build](https://travis-ci.org/github/aws/aws-xray-sdk-node/builds/739384919/config) passed and the only difference was disabling caching in the latter. The failed build was picking up an old version of a cached dependency.

However, I figure that has something to do with how lerna hoists dependencies so I'm sure the lerna-specific cache has it figured out. I'll add that.

anuraaga (approved):
Thanks!

PR #355, opened 2020-10-29
Fixes for GitHub Actions build

willarmiros:
*Issue #, if available:*
#352 

*Description of changes:*
This will likely be a long running PR to patch the GH actions config now that we can actually see the results of the patches in the CI. We can merge this once all tests are passing.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Checkout AWS XRay SDK Node Repository @ default branch latest
         uses: actions/checkout@v2
       - name: Setup Node ${{ matrix.node-version }}
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v1
         with:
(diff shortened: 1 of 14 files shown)


anuraaga (approved):

willarmiros:
Well that took a while. The first few fixes were with the GitHub actions config itself. Then Windows didn't like the Typescript definitions and importing/exporting syntax for the `postgres` and `mysql` packages. They weren't consistent with the ES6 syntax used in other packages anyway so that was a good fix to make. The final issue was that, when creating a test daemon in our integration tests, apparently Node 10 on Windows requires you to bind to `localhost` rather than `0.0.0.0` which worked on all other versions/runners. I added that as an answer [here](https://stackoverflow.com/questions/48733660/error-eaddrnotavail-on-socket-send-in-node-ssdp-protocol/64669279#64669279) in case any other poor soul ran into the same issue.

PR #361, opened 2020-12-15
Update Lambda trace IDs

willarmiros:
*Description of changes:*
Ensure that Lambda trace IDs will *never* be `null`, even outside the function handler. This extends #284. Also change the wording in the Lambda CME to clarify common cases that would cause the `_X_AMZN_TRACE_ID` to not be present, rather than simply stating it's not present.


lupengamzn (approved):
LGTM

Issue #360, opened 2020-12-15
Express patcher closes segment twice


willarmiros:
We have the possibility to close segments twice with the X-Ray Express middleware, once in the callback handler for the end of requests: https://github.com/aws/aws-xray-sdk-node/blob/a2344c47f60722b2b3b74eea5da897e86c09c5b4/packages/core/lib/middleware/mw_utils.js#L200

And again in the "after" express middleware: https://github.com/aws/aws-xray-sdk-node/blob/a2344c47f60722b2b3b74eea5da897e86c09c5b4/packages/express/lib/express_mw.js#L63-L72

We should investigate when this is possible and remove redundant closures.

willarmiros:
Released in 3.3.0

PR #362, opened 2020-12-21
Prevent Express faulted segments from being closed twice

willarmiros:
*Issue #, if available:*
#360

*Description of changes:*
Fixes a bug where, if there is a server fault for an express server, the segment would be closed twice. We had both the Express error-handling middleware and an event listener closing the segment before. The problem was that both places added important info: The error stack trace is only retrievable from the `closeSegment` middleware, and the HTTP response metadata is only available in the event listener. Before we were closing the segment with the error in the `closeSegment` middleware, now we are simply adding the error without closing. 

This didn't cause any behavioral impact to users because the X-Ray backend would be able to merge the data from both closed segments behind the scenes. However this should cause significant performance improvement since we no longer serialize twice, and it's less confusing to customers since we shouldn't be closing the same segment multiple times.

Added a unit and integration test to verify, and tested locally.

### DIFF
--- a/packages/express/lib/express_mw.js
+++ b/packages/express/lib/express_mw.js
@@ -1,3 +1,8 @@
+const AWSXRay = require('aws-xray-sdk-core');
+const { logger } = require('aws-xray-sdk-core/lib/middleware/sampling/service_connector');
+
+const mwUtils = AWSXRay.middleware;
+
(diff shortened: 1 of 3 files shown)


bhautikpip (approved):
LGTM :)

PR #363, opened 2020-12-22
Fix rowMode='array' in postgres patcher

### LINKED ISSUE #364: postgres patcher breaks rowMode: array query
Querying of the form 
```js
c.query({
  text: <query>,
  rowMode: 'array'
})
```

gets passed to pg incorrectly without the rowMode param.  I tried to fix in #363 but may have broken something.

isaacl:
The patcher incorrectly strips rowMode from calls of the form:
c.query({ text: <query>, rowMode: 'array' })

Fixes #364.

### DIFF
--- a/packages/postgres/test/unit/postgres_p.test.js
+++ b/packages/postgres/test/unit/postgres_p.test.js
@@ -56,7 +56,7 @@ describe('capturePostgres', function() {
       postgres.__query = function(args, values, callback) {
         this._queryable = true;
         this.queryQueue = [ null, null, queryObj ];
-        queryObj.callback = callback;
+        queryObj.callback = callback || (typeof args === 'object' ? args.callback : undefined);
         return queryObj;
       };
 
@@ -95,7 +95,7 @@ describe('capturePostgres', function() {
 
       sandbox.stub(AWSXRay, 'getNamespace').returns(session);
 
-      query.call(postgres, { sql: 'sql here', callback: function() {} });
+      query.call(postgres, { text: 'sql here', callback: function() {} });
       assert.equal(queryObj.callback.name, 'autoContext');
       queryObj.callback();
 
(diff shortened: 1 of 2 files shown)


isaacl:
hmm patch is causing tests to fail, idk I'll file an issue

isaacl:
I am still not sure this works, I had some weird behavior in xray but may be unrelated.

willarmiros:
Hi @isaacl,

Thanks for this contribution, do you think you could add a unit test to verify this fixes the problematic behavior?

isaacl:
OK I did a more involved patch where I use the argument logic from the pg module itself.

packages/postgres/test/unit/postgres_p.test.js
```diff
@@ -231,4 +231,65 @@ describe('capturePostgres', function() {
...
+      sandbox.stub(segment, 'addNewSubsegment').returns(subsegment);
+      sandbox.stub(AWSXRay, 'isAutomaticMode').returns(true);
+    });
+
+    afterEach(function() {
+      sandbox.restore();
+    });
+
+    it('should pass down raw sql query', function() {
+      query.call(postgres, 'sql here', ['values']);
```
willarmiros, line 283: Can we also assert that `addNewSubsegment` is `calledOnce` in both of these tests? Just to validate we are in fact correctly patching these queries in addition to propagating the params.

willarmiros (commented):
Looks great to me! This will be an awesome simplification, just one small comment. Thanks!

packages/postgres/test/unit/postgres_p.test.js
isaacl: sure, done

willarmiros (approved):
Thanks again! Verified with previous versions of the `pg` library that this change will be backwards compatible, since the `query()` API hasn't really changed apart from adding the query config object as an alternative to a string for the first argument, which this change adds support for.

isaacl:
can haz release?

PR #367, opened 2020-12-30
Wrap the catch method also

DavidTanner:
The current implementation of wrapping promises doesn't handle `new Promise().then(onSuccess).catch(onFailure)` setups.  ~~This is causing a circular dependency in ApolloServer.~~

~~I'm having trouble creating a test case for it, but am experiencing it when running in a Lambda function.~~

### DIFF
--- a/packages/core/test/unit/logger.test.js
+++ b/packages/core/test/unit/logger.test.js
@@ -1,18 +1,14 @@
-var expect = require('chai').expect;
-var rewire = require('rewire');
-var sinon = require('sinon');
-var logging = require('../../lib/logger');
+const expect = require('chai').expect;
+const rewire = require('rewire');
+const sinon = require('sinon');
+let logging = require('../../lib/logger');
 
 describe('logger', function () {
-  var spies = [];
+  const spies = [];
 
   function reloadLogger() {
-    var path = '../../lib/logger';
-    delete require.cache[require.resolve(path)];
-    logging = require(path);
-  }
-
-  before(function() {
+    const path = '../../lib/logger';
+    spies.length = 0;
     spies.push(sinon.spy(console, 'error'));
     spies.push(sinon.spy(console, 'warn'));
     spies.push(sinon.spy(console, 'info'));
@@ -21,19 +17,18 @@ describe('logger', function () {
     if (console.debug) {
       spies.push(sinon.spy(console, 'debug'));
     }
-  });
 
-  afterEach(function () {
-    spies.forEach(spy => spy.resetHistory());
-  });
+    delete require.cache[require.resolve(path)];
+    logging = require(path);
+  }
 
-  after(function () {
-    spies.forEach(spy => spy.restore());
+  afterEach(function() {
+    sinon.restore();
   });
 
   describe('console logging methods', function () {
-    before(function() {
-      process.env.AWS_XRAY_DEBUG_MODE = 'set';
+    beforeEach(function() {
+      process.env.AWS_XRAY_DEBUG_MODE = 'debug';
       reloadLogger();
     });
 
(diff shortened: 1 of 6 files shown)


srprash:
Hi @DavidTanner ! Thanks for identifying this gap. Patching the `catch` in addition to `then` is certainly required. 
I would like to understand your use case better. Would it possible to post a small snippet where you're running into this issue? This may help coming up with the test case for the PR as well. I didn't understand how not having the wrapper on catch can cause the circular dependency. Can you elaborate on that and how that appears only in Lambda function?

DavidTanner:
After debugging further I've found the circular dependency is caused by how exceptions are stored on the Segment, and specifically on the Lambda facade segment.

willarmiros:
@DavidTanner Exceptions are not meant to be stored on segments in Lambda, the Lambda facade segment available to your function code is immutable because the Lambda service manages the real segment that you ultimately see in the X-Ray console. Indeed, all the functions of the facade segment, including `addError`, are stubbed out to be no-ops: https://github.com/aws/aws-xray-sdk-node/blob/dd34e7e/packages/core/lib/env/aws_lambda.js#L51-L57

While I still think this is a valid change, I'm just a bit confused on where the "circular dependency" is.

DavidTanner:
See https://github.com/aws/aws-xray-sdk-node/pull/370 for a fix to the circular exception. 

The Error gets added to the facade segment, which will never clear it off because the methods are stubbed.  Using `capturePromise` causes the `error@context` to be added to the exception accross async methods.  In Apollo this error gets stringified and ends up as `ApolloException`.`originalError`.`error@context`.`segment`.`exception`.`ex` where `ex` is `originalError` thus creating the cycle that causes `JSON.stringify` to crash.

https://github.com/apollographql/apollo-server/blob/b7a91df76acef748488eedcfe998917173cff142/packages/apollo-server-core/src/runHttpQuery.ts#L108

https://github.com/apollographql/apollo-server/blob/b7a91df76acef748488eedcfe998917173cff142/packages/apollo-server-core/src/runHttpQuery.ts#L454

It has been brought up before for them to use a different method than JSON.stringify, but that has gone nowhere.

https://github.com/apollographql/apollo-server/issues/1433

willarmiros:
@DavidTanner thank you for the additional context. If you're willing, adding a small unit test to verify the CLS namespace is available in the catch block of a rejected promise would be appreciated. That being said, I understand we're currently lacking unit tests for this patcher so I'm willing to merge as is if you don't have the bandwidth.

DavidTanner:
I'll see what I can get together real quick @willarmiros

packages/core/test/unit/patchers/promise_p.test.js
```diff
@@ -0,0 +1,152 @@
...
+  describe('binds methods', () => {
+    beforeEach(() => {
+      getSegmentStub.returns({});
+      isAutomaticModeStub.returns(true);
+    });
+    createTests(true);
+  });
+});
+
+describe('capturePromise.patchThirdPartyPromise', () => {
```
willarmiros, line 130: I'm probably missing something, but where is this `patchThirdPartyPromise` function defined?

packages/core/test/unit/logger.test.js
```diff
@@ -162,113 +157,114 @@ describe('logger', function () {
...
 
     it('should not output if message and meta falsey', () => {
-      var logger = logging.getLogger();
+      const logger = logging.getLogger();
       logger.error();
       logger.error(null);
       logger.error('', null);
 
-      spies.forEach(spy => expect(spy).not.to.be.called);
+      spies.forEach(spy => sinon.assert.notCalled(spy));
```
willarmiros, line 176: Just for my knowledge, why replace some of these `expect`s with sinon assertions?

willarmiros (commented):
LGTM, and thanks @DavidTanner so much for adding these tests! Just a couple questions then good to merge.

packages/core/test/unit/patchers/promise_p.test.js
willarmiros: Oof well this was dumb, just didn't have the bottom 6 lines of `promise_p` loaded on the "files changed" view. This is good to go!

willarmiros (approved):

willarmiros:
I'll wait for @srprash to weigh in as well before merging.

srprash (approved):
Looks good to me too. Thanks for adding the tests and cleaning up other bits as well. :)

Issue #338, opened 2020-09-23
Node core dumps when run under rollup.


simonbuchan:
This is an odd one, I'm raising it here mostly as a signpost to what's going on for other people hitting it, since it's not really an AWS issue (really, it seems like a rollup or node issue), but possibly there's something this package can do here.

Basically, I use rollup to create my lambda deployment packages, and when I tried adding `aws-xray-sdk-core@3.2.0` it caused the lambda (runtime `nodejs12.x`) to fail with a core dump(!) with the message:

```
/var/lang/bin/node[7]: ../src/async_wrap.cc:308:void node::SetupHooks(const v8::FunctionCallbackInfo<v8::Value>&): Assertion `env->async_hooks_init_function().IsEmpty()' failed.
```

If I run the bundled .js file locally with node v14.10.0, I get pretty much the same error:

```
[19312]: c:\ws\src\async_wrap.cc:455: Assertion `env->async_hooks_init_function().IsEmpty()' failed.
```

With some experimenting, I've narrowed it down to simply importing `async-hook-jl`, which is imported by `cls-hooked` when it detects node is less than v8. Due to the way rollup works, this is actually always included and run. I'm not sure if there's a better option, but I got this to work under rollup with the following mini-plugin, which also disables the `pkginfo` module that's throwing under rollup too:

```js
{
  name: "exclude-modules",
  files: [
    "cls-hooked/context-legacy",
    "pkginfo",
  ].map((id) => require.resolve(id)),
  load(file) {
    if (this.files.includes(file)) {
      return { code: "" };
    }
  },
},
```

Ideally, since node <10 is EOL, it would be nice if the `cls-hooked` dependency could be dropped and the native `async_hooks` module was used directly, or a more modern replacement used, which would avoid this particular headache, but since you've just updated to it to drop node <8 support, I'm guessing that's not likely.

willarmiros:
Hi @simonbuchan,
Thanks for reporting this. I'm not too familiar with rollup but this:

> which is imported by cls-hooked when it detects node is less than v8. Due to the way rollup works, this is actually always included and run

does not sound like desirable behavior, is that a bug that you could open with Rollup?

[...]

PR #369, opened 2020-12-31
Better bundling

### LINKED ISSUE #368: Remove pkginfo dependency
The pkginfo module causes issues with aws-xray-sdk-core being bundled by tools like rollup. This is described in issue #338. Most of the discussion in that issue centered around cls-hooked which by far causes bigger issues (Node core dumping) however pkginfo also breaks when it's used outside of a standard node environment.

Usage of pkginfo is limited to an invocation in the aws-xray-sdk-core (https://github.com/aws/aws-xray-sdk-node/blob/dd34e7e2c6f4267123707e166d2320d4bdfd14b3/packages/core/lib/aws-xray.js#L9) and a repeated invocation in the full sdk https://github.com/aws/aws-xray-sdk-node/blob/dd34e7e2c6f4267123707e166d2320d4bdfd14b3/packages/full_sdk/lib/index.js#L8

In both cases the the invocation of pkginfo could be wrapped in a try catch block to simply suppress errors (the pkginfo module is already skipped when moduel.filename is undefined) or the code could be replaced with an import of the package.json file which we know the relative path to. Additionally any fields that are being set by pkginfo should be called out in the .d.ts file for these modules.

chris-pardy:
Resolves #368 and kind-of #338 

*Description of changes:*
The cls-hooked and pkginfo dependencies are problematic for bundlers like rollup (see #338) to make this work better we can do 2 simple things:

1. For cls-hooked importing `cls-hooked/context` directly rather than using the compatibility layer at the package index.js file has no performance or behavior implication and simply skips a check that this is being run in node versions greater than or equal to 8 (current min engine version is 10). Ideally we'd be able to ditch the cls-hooked dependency entirely but that is a much larger change.
2. For pkginfo the package simply climbs the file path starting from a module until it finds a `package.json` file, since we know where our package.json file will be it's easy to just import the values from that file. This will resolve #368 

### DIFF
--- a/packages/core/lib/aws-xray.js
+++ b/packages/core/lib/aws-xray.js
@@ -4,10 +4,17 @@ var segmentUtils = require('./segments/segment_utils');
 var utils = require('./utils');
 var LambdaEnv = require('./env/aws_lambda');
 
-var UNKNOWN = 'unknown';
+// Import Data from package.json,
+// If the importing of package.json fails leave
+// pkginfo as an empty object
+var pkginfo = {}
+try {
+  pkginfo = require('../package.json');
+} catch (err) {
+  logging.getLogger().debug('Failed to load SDK data:', err);
+}
 
-var pkginfo = module.filename ? require('pkginfo') : function() {};
-pkginfo(module);
+var UNKNOWN = 'unknown';
 
 /**
  * A module representing the AWSXRay SDK.
@@ -351,8 +358,8 @@ AWSXRay.middleware.IncomingRequestData = require('./middleware/incoming_request_
 
   var sdkData = {
     sdk: 'X-Ray for Node.js',
-    sdk_version: (module.exports && module.exports.version) ? module.exports.version : UNKNOWN,
-    package: (module.exports && module.exports.name) ? module.exports.name : UNKNOWN,
+    sdk_version: pkginfo.version ? pkginfo.version : UNKNOWN,
+    package: pkginfo.name ? pkginfo.name : UNKNOWN,
   };
 
   segmentUtils.setSDKData(sdkData);
--- a/packages/core/test/unit/aws-xray.test.js
+++ b/packages/core/test/unit/aws-xray.test.js
@@ -44,7 +44,17 @@ describe('AWSXRay', function() {
 
       assert.property(setSDKDataStub.firstCall.args[0], 'sdk');
       assert.property(setSDKDataStub.firstCall.args[0], 'sdk_version');
+      assert.notStrictEqual(
+        setSDKDataStub.firstCall.args[0].sdk_version,
+        'unknown',
+        'Expected sdk_version to not be unknown'
+      );
       assert.property(setSDKDataStub.firstCall.args[0], 'package');
+      assert.notStrictEqual(
+        setSDKDataStub.firstCall.args[0].package,
+        'unknown',
+        'Expected package to not be unknown'
+      );
 
       assert.property(setServiceDataStub.firstCall.args[0], 'runtime');
       assert.property(setServiceDataStub.firstCall.args[0], 'runtime_version');
--- a/packages/full_sdk/lib/index.js
+++ b/packages/full_sdk/lib/index.js
@@ -4,14 +4,22 @@ AWSXRay.express = require('aws-xray-sdk-express');
 AWSXRay.captureMySQL = require('aws-xray-sdk-mysql');
 AWSXRay.capturePostgres = require('aws-xray-sdk-postgres');
 
+// Import Data from package.json,
+// If the importing of package.json fails leave
+// pkginfo as an empty object
+var pkginfo = {}
+try {
+  pkginfo = require('../package.json');
+} catch (err) {
+  AWSXRay.getLogger().debug('Failed to load SDK data:', err);
+}
+
 var UNKNOWN = 'unknown';
-var pkginfo = module.filename ? require('pkginfo') : function() {};
-pkginfo(module);
 
 (function () {
   var sdkData = AWSXRay.SegmentUtils.sdkData || { sdk: 'X-Ray for Node.js' };
-  sdkData.sdk_version = (module.exports && module.exports.version) ? module.exports.version : UNKNOWN;
-  sdkData.package = (module.exports && module.exports.name) ? module.exports.name : UNKNOWN;
+  sdkData.sdk_version = pkginfo.version ? pkginfo.version : UNKNOWN;
+  sdkData.package = pkginfo.name ? pkginfo.name : UNKNOWN;
   AWSXRay.SegmentUtils.setSDKData(sdkData);
 })();
 
--- a/packages/core/lib/context_utils.js
+++ b/packages/core/lib/context_utils.js
@@ -2,7 +2,7 @@
  * @module context_utils
  */
 
-var cls = require('cls-hooked');
+var cls = require('cls-hooked/context');
 
 var logger = require('./logger');
 var Segment = require('./segments/segment');
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -19,7 +19,6 @@
     "@types/cls-hooked": "*",
     "atomic-batcher": "^1.0.2",
     "cls-hooked": "^4.2.2",
-    "pkginfo": "^0.4.0",
     "semver": "^5.3.0"
   },
   "scripts": {
--- a/packages/full_sdk/package.json
+++ b/packages/full_sdk/package.json
@@ -16,8 +16,7 @@
     "aws-xray-sdk-core": "file:../core",
     "aws-xray-sdk-express": "file:../express",
     "aws-xray-sdk-mysql": "file:../mysql",
-    "aws-xray-sdk-postgres": "file:../postgres",
-    "pkginfo": "^0.4.0"
+    "aws-xray-sdk-postgres": "file:../postgres"
   },
   "scripts": {
     "test": "tsd",
--- a/packages/test_express/test/helpers.js
+++ b/packages/test_express/test/helpers.js
@@ -2,7 +2,7 @@ var dgram = require('dgram');
 var xray = require('aws-xray-sdk-core');
 var xrayExpress = require('aws-xray-sdk-express');
 var express = require('express');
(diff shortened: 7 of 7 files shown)


willarmiros (commented):
1. I'm ok with the `cls-hooked` change since we're just shortcutting the behavior that `cls-hooked` is already doing under the hood. It is somewhat depending on another package's internal behavior, but given the staleness of `cls-hooked` I don't think it's bound to change.
2. Tested this on my sample app and it did not maintain the same behavior.

Snippet of my segment before:

```
                "aws": {
                    "xray": {
                        "package": "aws-xray-sdk-core",
                        "sdk_version": "3.2.0",
                        "sdk": "X-Ray for Node.js"
                    }
                },
```

Snippet of my segment after:

```
               "aws": {
                    "xray": {
                        "package": "unknown",
                        "sdk_version": "unknown",
                        "sdk": "X-Ray for Node.js"
                    }
                },
```

I actually don't see where `module.exports` is getting overwritten as you say. The call to `pkginfo(module)` stores the SDK data in `module.exports`, and then that data is read from `module.exports` [here](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/aws-xray.js#L354-L355). We do later [overwrite](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/aws-xray.js#L365) `module.exports`, but that's after the data is read.

chris-pardy:
> 1. I'm ok with the `cls-hooked` change since we're just shortcutting the behavior that `cls-hooked` is already doing under the hood. It is somewhat depending on another package's internal behavior, but given the staleness of `cls-hooked` I don't think it's bound to change.
> 2. Tested this on my sample app and it did not maintain the same behavior.
> 
> Snippet of my segment before:
> 
> ```
>                 "aws": {
>                     "xray": {
>                         "package": "aws-xray-sdk-core",
>                         "sdk_version": "3.2.0",
>                         "sdk": "X-Ray for Node.js"
>                     }
>                 },
> ```
> 
> Snippet of my segment after:
> 
> ```
>                "aws": {
>                     "xray": {
>                         "package": "unknown",
>                         "sdk_version": "unknown",
>                         "sdk": "X-Ray for Node.js"
>                     }
>                 },
> ```
> 
> I actually don't see where `module.exports` is getting overwritten as you say. The call to `pkginfo(module)` stores the SDK data in `module.exports`, and then that data is read from `module.exports` [here](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/aws-xray.js#L354-L355). We do later [overwrite](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/aws-xray.js#L365) `module.exports`, but that's after the data is read.

Yep, I missed where it was read then overwritten. I was thinking that the data would be exported. I'll add a test for those values and then remove the pkginfo dependency.

chris-pardy:
@willarmiros Should be working now.

willarmiros (commented):
@chris-pardy yep it's working on my end! Just a couple comments. You also mentioned adding a test, which would be awesome if you're able!

chris-pardy:
> @chris-pardy yep it's working on my end! Just a couple comments. You also mentioned adding a test, which would be awesome if you're able!

@willarmiros  There was an existing test that the value was set to something. I added asserts to ensure that it wasn't set to unknown which should catch the issue I had originally introduced, without also requiring the tests to update anytime the package version get's bumped.

packages/core/test/unit/aws-xray.test.js
```diff
@@ -44,7 +44,17 @@ describe('AWSXRay', function() {
...
+        setSDKDataStub.firstCall.args[0].sdk_version,
+        'unknown',
+        'Expected sdk_version to not be unknown'
+      );
       assert.property(setSDKDataStub.firstCall.args[0], 'package');
+      assert.notStrictEqual(
+        setSDKDataStub.firstCall.args[0].package,
+        'unknown',
+        'Expected package to not be unknown'
+      );
```
chris-pardy, line 57: @willarmiros I'm not super familiar with chai and didn't spend that long googling. I know in Jest I could do something that would assert the structure of this object and the values of it's key's in one shot, I'm not sure if there's an equivalent in chai or if this approach of independent asserts is prefered.

packages/core/lib/aws-xray.js
```diff
@@ -3,16 +3,7 @@ var logging = require('./logger');
...
-// Import Data from package.json,
-// If the importing of package.json fails leave
-// pkginfo as an empty object
-var pkginfo = {}
-try {
-  pkginfo = require('../package.json');
-} catch (err) {
-  logging.getLogger().debug('Failed to load SDK data:', err);
-}
+var pkginfo = module.filename ? require('../package.json') : {};
```
chris-pardy, line 6: @willarmiros this change is perhaps a little closer to the behavior that was in place originally. I believe the `module.filename` check is there for something like webpack which doesn't play well with pgkinfo (but in a less crash and burn on startup way then a bundler like rollup).
However the other thing I've been thinking about is if this change is safe as a point release. In the case of bundlers this change may break because importing a json file will usually require a json loader which may or may-not be configured. However if the json loader is present we can start getting the xray sdk data loaded when previously it would have been unknown. I'm happy to defer to you in terms of which change makes the most sense.

1) require package.json in a try/catch and log an error if it can't be required - Probably a result of bundling
2) skip the require call if modile.filename is falsey - probably turns off require call for some bundlers
3) Unconditional require (my preference) - AWS-SDK won't be usable if the require fails but the code is the cleanest and the failure cases involve running in non-node environments.

packages/core/test/unit/aws-xray.test.js
willarmiros: You can probably use [hasAllKeys](https://www.chaijs.com/api/assert/#method_hasallkeys) instead of asserting for individual properties how we're doing it now, but as for checking that they're not unknown the way you're doing it now seems fine.

packages/core/lib/aws-xray.js
willarmiros: Reading through [the source](https://github.com/indexzero/node-pkginfo/blob/master/lib/pkginfo.js), pkginfo is basically just requiring `../package.json` exactly as you are now. Though I saw their `require` is [wrapped by try/catch](https://github.com/indexzero/node-pkginfo/blob/master/lib/pkginfo.js#L103-L105), so I was wondering how this breaks in a non-standard environment as you said in #368?

Regardless, I would not be in favor of option 3, because even if we held this change til a major version bump making the SDK unusable for existing customers simply because they bundle their `package.json` unorthodoxly is silly. I'd be in favor of option 1, but at that point it seems like we're reinventing the wheel and just doing exactly the work of `pkginfo` ourselves.

willarmiros (commented):
We'll probably need a bit more discussion on this still.

packages/core/lib/aws-xray.js
chris-pardy: @willarmiros I'm happy to leave the version with the try/catch/log setup. I wrote up some notes on the failure case on the [original issue](https://github.com/aws/aws-xray-sdk-node/issues/368#issuecomment-769348677). I'll drop the other commits off this PR and build a test setup where I can see what will happen if a bundler is not setup to load json and this pr where to be applied. I'll add what I find to this issue.

As for replicating the work of pkginfo ourselves directly requiring the package.json from a known location is a pretty substantial improvement. The `find` code in pkginfo does several directory checks, directory traversals and tries to require a package.json file. From a performance standpoint this library will require substantially less file access/read calls with the direct import. Beyond that a properly configured bundler that can load json will now be able to correctly include the x-ray-sdk data in it's payloads.

chris-pardy:
@willarmiros 
> We'll probably need a bit more discussion on this still.

I've done some more digging. It turns out there are already json files being required by this library [here](https://github.com/aws/aws-xray-sdk-node/tree/master/packages/core/lib/resources) This means that anyone using this code will already have a solution in pace for handling json files. While it is possible that someone could be doing something really exotic that allowed this code to work with pkginfo but would now break with the direct package.json require I'm not sure if this package needs to account for that when versioning. I think the import with the try/catch is probably the safest option, however I think just a straight require call could easily be considered a patch change..

The other potentially "breaking" change is dropping the pkginfo dependency. By doing so we could break code that depended on that transitive dependency (eg. using pkginfo without declaring the dependency and letting aws-xray-sdk bring it along for the ride) Again I'm not sure what the policy in terms of breaking changes are, but I would not count this as one.

packages/core/test/unit/aws-xray.test.js
chris-pardy: going to leave this as-is then.

packages/core/lib/aws-xray.js
willarmiros: Ok I'm sold on the idea that this require statement is an improvement on using `pkginfo`.

willarmiros:
> While it is possible that someone could be doing something really exotic that allowed this code to work with pkginfo

I agree, it seems to me like the only case this breaks is if you somehow bundled this SDK such that its `package.json` is in some ancestor directory as opposed to its parent directory, which is a narrow enough case. Regardless, I see no reason to not be safe where we can and use the try/catch as you are currently. This way it would definitely not be considered a breaking change, we've removed dependencies on minor version changes in the past.

packages/full_sdk/lib/index.js
```diff
@@ -4,14 +4,22 @@ AWSXRay.express = require('aws-xray-sdk-express');
...
 AWSXRay.capturePostgres = require('aws-xray-sdk-postgres');
 
+// Import Data from package.json,
+// If the importing of package.json fails leave
+// pkginfo as an empty object
+var pkginfo = {}
+try {
+  pkginfo = require('../package.json');
+} catch (err) {
+  logging.getLogger().debug('Failed to load SDK data:', err);
```
willarmiros, line 14:
```suggestion
  AWSXRay.getLogger().debug('Failed to load SDK data:', err);
```

willarmiros (commented):
Just the suggested change then good to merge! I'll hold off til you commit it in case you want to mention anything else @chris-pardy

chris-pardy:
@willarmiros Good to go! Thanks for the Review!

willarmiros (approved):

PR #370, opened 2021-01-04
Use addError to set parent exceptions instead of .exception

DavidTanner:
Currently when an exception is added to a subsegment the `Segment.exception` doesn't get cleaned up.  This causes errors when internal services try to stringify the exception.

### DIFF
--- a/packages/core/.eslintrc.json
+++ b/packages/core/.eslintrc.json
@@ -29,6 +29,6 @@
         ]
     },
     "parserOptions": {
-        "ecmaVersion": 6
+        "ecmaVersion": 2019
     }
 }
--- a/packages/core/lib/segments/attributes/subsegment.js
+++ b/packages/core/lib/segments/attributes/subsegment.js
@@ -182,20 +182,8 @@ Subsegment.prototype.addError = function addError(err, remote) {
 
   this.addFaultFlag();
 
-  if (this.segment && this.segment.exception) {
-    if (err === this.segment.exception.ex) {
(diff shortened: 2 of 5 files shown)


DavidTanner:
This addresses things like https://github.com/apollographql/apollo-server/issues/1433 on lambda where the Lambda XRay facade doesn't clean up the `.exception` value that the `Subsegment` assigns to it.

willarmiros:
@DavidTanner this appears to be a problem unique to Lambda and the facade segment, is that correct? While I agree that the `segment.exception` not being cleaned up is a bug, perhaps we should be cleaning it up in the facade segment's [reset function](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/env/aws_lambda.js#L71-L76) which is called at the start of each invocation. 

I'm just having a bit of a hard time reasoning about whether this change affects the behavior of `addError`, namely in [this case](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/segments/attributes/subsegment.js#L185) where `segment.exception` is already defined and `subsegment.addError` will point its `cause` to the ID of the captured exception if the new exception is the same as the existing one.

DavidTanner:
@willarmiros in that case I think the right thing to do would be to check out the exception is in the segment.  I can update the code to check for that.

NathanielRN:
Hey @DavidTanner ! Are you still interested in working on this PR? If you are please let us know when it's ready for a review 🙂

PR #375, opened 2021-01-28
Make @types/cls-hooked a devDependency

willarmiros:
*Description of changes:*
Type defs are not needed at runtime and should therefore be dev dependencies.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -16,7 +16,6 @@
     "test": "test"
   },
   "dependencies": {
-    "@types/cls-hooked": "*",
     "atomic-batcher": "^1.0.2",
     "cls-hooked": "^4.2.2",
     "pkginfo": "^0.4.0",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -15,7 +15,9 @@
   "directories": {
     "test": "test"
   },
+  "//": "@types/cls-hooked is exposed in API so must be in dependencies, not devDependencies",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


DavidTanner:
From personal experience I recommend keeping types that are used in the published library in the dependency list. Otherwise downstream typescript projects have to include the types.

anuraaga:
Yeah looks like we export `cls` types in our APIs

https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/context_utils.d.ts#L1

If it was a pure implementation detail, `devDependencies` for the types would be no problem, but when in the API surface I guess we should include in `dependencies`.

packages/core/package.json
```diff
@@ -16,7 +16,6 @@
     "test": "test"
   },
   "dependencies": {
```
anuraaga, line 18: But someone will get confused again in a year or two :) It's too bad package.json comments are terrible, but this might work

```suggestion
  "dependencies": {
     "//": "@types/cls-hooked exposed in API so can't be in devDependencies",
```

willarmiros: I guess you can't nest comments in the `dependencies` block according to these comments https://stackoverflow.com/a/14221781/8185365. I'll add the comment above depdencies

willarmiros:
@DavidTanner @anuraaga got it, that definitely makes sense as to why it was a devDependency and the other `@types/*` packages weren't. I'll add a comment indicating such for the future.

anuraaga (approved):

PR #377, opened 2021-02-15
Add package lock to repo

willarmiros:
*Description of changes:*
Previously, we did not commit the `package-lock.json` file because it caused many headaches with merge conflicts and lerna builds. But now that we've cleaned up our hoisted dependencies with lerna so there shouldn't be build problems, and the benefits of reproducible builds should outweigh the occasional conflict.


srprash:
How is the package-lock.json generated? Also, in the CI workflow the dependencies are resolved during the build. How will the package-lock.json be updated in that case?

willarmiros:
package-lock is created/updated by running `lerna bootstrap --hoist`, that is the default behavior that we had disabled before via the `--no-package-lock` argument in lerna.json. 

> in the CI workflow the dependencies are resolved during the build. How will the package-lock.json be updated in that case?

In the CI workflow (where `CI=true`), lerna uses `npm ci` under the hood which reads the package-lock.json to determine what dependencies to use per the [lerna docs](https://github.com/lerna/lerna/tree/main/commands/bootstrap#--no-ci). This way we have a reproducible build with deterministic dependencies. It's a read-only operation, there's no need to update it.

srprash:
I see. Does that mean that once the `package-lock.json` file is created/updated, it will be used for every build unless modified manually? I think more than a reproducible build it guarantees repeatable build where we know exactly what depencies are bing pulled in.

anuraaga:
@willarmiros Looks like the workflow explicitly has `npm install` - maybe we can remove it? Or otherwise need to rename to `npm ci`

willarmiros:
@srprash the package-lock will be strictly read for every build run using `npm ci`. For builds using `npm install` the package.json will be used, and package-lock will be updated automatically if necessary (e.g. if a new dependency is added). The package-lock file should never be updated manually; if someone were making a PR that added or upgraded a dependency and needed package-lock to be updated, they'd run `npm install` locally to update it and include the update in their PR.

@anuraaga yep good catch, it's safe to remove and was only there because we weren't hoisting correctly before.

willarmiros:
For `npm ci` you want to cache the `~/.npm` local repository of packages instead of `node_modules`, presumably to reuse dependencies from the local repo instead of fetching them for all packages.

.github/workflows/pr-build.yml
```diff
@@ -21,23 +21,24 @@ jobs:
...
           node-version: ${{ matrix.node-version }}
+
       - name: Cache NPM modules
         uses: actions/cache@v2
         with:
-          path: |
-            node_modules
-            */*/node_modules
-          key: ${{ matrix.os }}-${{ hashFiles('**/package.json') }}
+          path: $HOME/.npm
```
anuraaga, line 37: This seems natural to me for sure, but any idea why github actions would recommend node_modules?

https://github.com/actions/cache/blob/main/examples.md#node---lerna

srprash (approved):
I'm not familiar with the npm ci stuff but for the package-lock usage and update mechanism looks good.

anuraaga (approved):

PR #378, opened 2021-02-22
Add Code Coverage

willarmiros:
*Description of changes:*
Adds code coverage reporting to repo for PRs and master branch builds.


willarmiros:
Ok, this should be ready for review. The commits are being reported to codecov here: https://app.codecov.io/gh/aws/aws-xray-sdk-node/commits. This is similar to the OpenTelemetry JS Contrib's [coverage page](https://app.codecov.io/gh/open-telemetry/opentelemetry-js-contrib/commits?page=1). We do have the issue (both here and on the Java SDK's codecov) of commits not being found, but it looks like that's something we can resolve once the aws org adds the app to our repos according to [this](https://community.codecov.io/t/unable-to-find-commit-in-github-in-the-recent-dashboard/1145/5).

I was also initially worried because the page for this PR (https://app.codecov.io/gh/aws/aws-xray-sdk-node/compare/378) was a 404, but it turns out that's because there's no coverage reports on master to compare it to. The only thing I'm not sure about is the codecov bot commenting the diff diagram, but I think that'll happen if we give it a few hours.

.github/workflows/pr-build.yml
```diff
@@ -21,6 +21,10 @@ jobs:
           - 12.x
           - 14.x
           - 15.x
+        include:
+          - os: ubuntu-latest
+            node-version: 14.x
```
srprash, line 26: Why node 14.x and not 15.x?

srprash (approved):
Looks good.
Had the same issue with [aws-xray-java-agent](https://app.codecov.io/gh/aws/aws-xray-java-agent) where codecov not unable to find the commit which disappeared without even adding the codecov app to the repo. This intermittent issue is reported on the discussion you mentioned. I guess its codecov just being flaky.

anuraaga (approved):

.github/workflows/pr-build.yml
willarmiros: Odd numbered major versions of node are effectively permanent RCs for the next "real" even-numbered major version release. The odd-numbered versions are only supported for 6 months, and generally not used in production. You can read more here: https://nodejs.org/en/about/releases/

PR #397, opened 2021-04-01
Add smoke test of distribution channel

lupengamzn:
*Issue #, if available:*

*Description of changes:*
Added smoke test of distribution channel

### DIFF
--- a/.github/workflows/continuous-monitoring.yml
+++ b/.github/workflows/continuous-monitoring.yml
@@ -2,8 +2,8 @@ name: Continuous monitoring of distribution channels
 
 on:
   workflow_dispatch:
-    schedule:
-      - cron: '0 */1 * * *'
+  schedule:
+    - cron: '0 */1 * * *'
 
 jobs:
   smoke-tests:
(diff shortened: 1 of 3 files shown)


.github/workflows/continuous-monitoring.yml
```diff
@@ -0,0 +1,39 @@
...
+    - cron: '0 */1 * * *'
+
+jobs:
+  smoke-tests:
+    name: Run smoke tests for Node ${{ matrix.node-version }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os:
+          - macos-latest
```
bhautikpip, line 15: think we can just keep `ubuntu-latest` and latest version of node should be good enough.

bhautikpip (approved):
Just one minor comment. Otherwise looks good :)

.github/workflows/continuous-monitoring.yml
lupengamzn: Just updated

Issue #398, opened 2021-04-02
`capturePromises` is not idempotent


fossamagna:
The problem with calling things like `capturePromises` is that it is adding another function to the `.then`.  Depending on how many calls a lambda gets, it could end up adding a lot of functions to the Promise prototype slowing things down.

```javascript
function patchPromise(Promise) {
  var then = Promise.prototype.then;
  Promise.prototype.then = function(onFulfilled, onRejected) {
    if (contextUtils.isAutomaticMode()
      && tryGetCurrentSegment()
    ) {
      var ns = contextUtils.getNamespace();

      onFulfilled = onFulfilled && ns.bind(onFulfilled);
      onRejected = onRejected && ns.bind(onRejected);
    }

    return then.call(this, onFulfilled, onRejected);
  };
}
```
https://github.com/aws/aws-xray-sdk-node/blob/aws-xray-sdk%403.1.0/packages/core/lib/patchers/promise_p.js#L12-L26

_Originally posted by @DavidTanner in https://github.com/aws/aws-xray-sdk-node/issues/143#issuecomment-754744567_

PR #400, opened 2021-04-02
Make it idempotent to capturePromise

fossamagna:
*Issue #, if available:*
#398

*Description of changes:*
Make it idempotent to capturePromise.
Add uncapturePromise function in order to cancel patch in tests.

### DIFF
--- a/packages/core/test/integration/segment_maintained_across_shared_promise.test.js
+++ b/packages/core/test/integration/segment_maintained_across_shared_promise.test.js
@@ -37,6 +37,8 @@ var server = http
         // setTimeout so the assertion isn't caught by the promise
         setTimeout(function() {
           assert.equal(segment.id, retrievedSegment.id);
+          // Cancel the patch because it doesn't affect other tests
+          require('../../lib/patchers/promise_p').uncapturePromise();
         });
       });
     });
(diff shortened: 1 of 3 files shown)


packages/core/test/integration/segment_maintained_across_shared_promise.test.js
```diff
@@ -70,3 +70,6 @@ function sendRequest(address, cb) {
     .on('error', cb)
     .end();
 }
+
+// Cancel the patch because it doesn't affect other tests
+require('../../lib/patchers/promise_p').uncapturePromise();
```
willarmiros, line 75: So this is effectively cleanup for the integration test? If so, can we add it after the assertion so we're confident it's executed after the test will execute? Or perhaps wrap things in a `try..finally` block? Please inform me if I'm misunderstanding the purpose here.

willarmiros (commented):
Thanks for the contribution! Generally lgtm, just one comment/question.

packages/core/test/integration/segment_maintained_across_shared_promise.test.js
fossamagna: @willarmiros Thank you for your review. I moved cleanup for integration test after assertion, and change setup and cleanup process in `promise_p.test.js` too.

willarmiros (approved):
Thanks!

Issue #385, opened 2021-03-02
[Bug] SDK emit content_length in http response with string type instead of number.


JohnWu20:
Xray SDK should send `content_length` with number type in the json segment for the [http request data](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html). 

But I get this data when I use the latest node.js sdk version: [data.txt](https://github.com/aws/aws-xray-sdk-node/files/6065846/data.txt)

This is a bug and when our customers try to use OT collector to take place of Xray-daemon, they can't get the data they want.

awssandra:
This looks correct, as per https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html content-length should be recorded as a number.

We'll need to correct this in a few places:
https://github.com/aws/aws-xray-sdk-node/search?q=content_length

PR #402, opened 2021-04-06
fixed content_length and status always ints

willarmiros:
*Issue #, if available:*
#385

*Description of changes:*
Ensures `content_length` and `status` attributes are always integers, as specified in [our docs](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html#api-segmentdocuments-http). Also did some refactoring to reduce repeated code.

### DIFF
--- a/packages/core/lib/segments/segment_utils.js
+++ b/packages/core/lib/segments/segment_utils.js
@@ -1,3 +1,4 @@
+const { safeParseInt } = require('../utils');
 var logger = require('../logger');
 
 var DEFAULT_STREAMING_THRESHOLD = 100;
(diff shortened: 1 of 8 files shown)


packages/core/lib/segments/segment_utils.js
```diff
@@ -45,6 +46,27 @@ var utils = {
...
+    if (!res) {
+      return ret;
+    }
+    
+    const status = safeParseInt(res.statusCode);
+    if (status !== 0) {
+      ret.status = status;
+    }
+    if (res.headers && res.headers['content-length']) {
+      ret.content_length = safeParseInt(res.headers['content-length']);
```
lupengamzn, line 67: Should we keep recording the `content-length` even if it's `0`?

willarmiros: Yes we should IMO, `content-length` can be 0 if there is a response with an empty body, which could still be valuable for a customer to know.

lupengamzn (approved):
LGTM

PR #403, opened 2021-04-06
Deprecate package-specific changelogs

willarmiros:
*Description of changes:*
We should store all changelog entries for the entire repo in a single CHANGELOG file at the root of the repo, as is standard practice for monorepos, to avoid confusion. This PR adds a note that future CHANGELOG entries will only be added to the top-level CHANGELOG.md file.


bhautikpip (approved):
Should we also add this package to the list as well (https://github.com/aws/aws-xray-sdk-node/blob/master/packages/full_sdk/CHANGELOG.md)?

willarmiros:
The `full_sdk` package doesn't actually contain any business logic, it just depends on the core, express, mysql, and postgres packages and *used* to contain the main changelog for all the other packages. If there is ever any substantive changes or actual business logic added to the `full_sdk` we can start including it in the changelog, but for now I think it's ok as is.

Issue #411, opened 2021-04-12
After installing sdk version 3.3.0 Typescript compilation error


organic-scholar:
../node_modules/aws-xray-sdk-core/lib/patchers/aws3_p.ts:7:40 - error TS2307: Cannot find module '@aws-sdk/config-resolver' or its corresponding type declarations.

7 import type { RegionInputConfig } from '@aws-sdk/config-resolver';
                                         ~~~~~~~~~~~~~~~~~~~~~~~~~~

../node_modules/aws-xray-sdk-core/lib/patchers/aws3_p.ts:85:43 - error TS2339: Property 'serviceId' does not exist on type 'Configuration'.

85   const serviceIdentifier = client.config.serviceId;
                                             ~~~~~~~~~


Found 2 errors.


The typescript compiler is giving the above errors on build.

Dema1348:
I have the same error now compiling a project in typescript

awssandra:
Apologies for the issue! We're taking a look.

awssandra:
Hey everyone,

Does anyone have a clean reproduction case?  Or, additional stack traces would be beneficial.

willarmiros:
@organic-scholar @Dema1348 I'm attempting to reproduce this issue with a TypeScript sample app but am having no luck. I'm using v3.7.3 of `tsc`, v9.1.1 of `ts-node` to run my sample TypeScript app. My `tsconfig.json` is:

```
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "es5",
    "strict": true,
    "strictNullChecks": true,
    "declaration": true,
    "esModuleInterop": true
  },
  "files": ["index.ts"]
}
```

Can you provide the details of your setup requested by @awssandra so we can try to reproduce this locally?

willarmiros:
@organic-scholar @Dema1348 I've released v3.3.1 which rolls back the AWS SDK V3 support and should resolve your compiler issues for now while we work to resolve the underlying issue. Please let me know if upgrading to 3.3.1 fixes this problem.

organic-scholar:
Thanks, @willarmiros v3.3.1 compiled successfully, 

my project has the following ts version and config

typescript: ~3.9.7

```
{
  "compilerOptions": {
    "target": "ES2018",
    "module": "CommonJS",
    "lib": ["es2018"],
    "declaration": false,
    "strict": true,
    "noImplicitAny": true,
    "strictNullChecks": true,
    "noImplicitThis": true,
[...]

PR #386, opened 2021-03-06
refs #294, add support for AWS SDK v3

monken:
#294

* add support for AWS SDK v3

I used typescript for the actual patcher because it makes the code safer and I wanted to make sure that types are maintained when using the XRay patcher. I included the complied JavaScript code as well so that tests can be run without adding more typescript flavor to the project.

There is no captureAWS for now and I also haven't looked at manual mode much because I'm unsure about how to pass in the xray params in a type-safe way.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #414, opened 2021-04-13
Add back AWS SDK V3 support

willarmiros:
*Issue #, if available:*
#294

*Description of changes:*
This PR does absolutely nothing except undoing the revert commit that removed #386. So this PR is identical to #386. Future fixes will be needed to address #411 and some other issues, but I think it is best to apply those fixes on an individual basis, rather than grouping them with this behemoth change. 

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,119 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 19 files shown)


willarmiros:
Will work on a separate branch (https://github.com/aws/aws-xray-sdk-node/tree/aws-v3-support) to keep working tree clean while we get to a working state.

PR #416, opened 2021-05-05
Remove type

willarmiros:
*Issue #, if available:*
#411

*Description of changes:*
This change redesigns the AWS SDK V3 client instrumentation to use the AWS SDK's `Pluggable` interface, which makes all the instrumentation middleware-based instead of a hybrid solution of middleware and monkey-patching. There was an original plan to vend solely a plugin, which customers could use like:

```js
s3.middlewareStack.use(AWSXRay.getXRayPlugin());
```

However, I decided to stick with the original user experience of:

```js
const s3 = new S3({});
AWSXRay.captureAWSv3Client(s3);
```

This is for 3 reasons:

1. The former was very difficult to make work with manual mode, where a Segment must be manually passed in to the plugin by the user. I was constantly running into this issue when trying to test it with a sample app: https://github.com/aws/aws-sdk-js-v3/issues/1803
2. The latter experience is more idiomatic for X-Ray users, compared to how instrumentation was done for AWS SDK v2 and all other AWS SDKs across languages
2. The latter is more extensible to OpenTelemetry, which requires an Otel-based API to be used for instrumentation as opposed to using a library's API (like `middlewareStack`)

I was still unable to get the latter API to work perfectly with TypeScript, because I would still get the error described in https://github.com/aws/aws-sdk-js-v3/issues/1803. But a workaround was to just cast the client to `any` while instrumenting, which is not ideal, but worked once I did it.

```ts
const s3 = new S3({});
AWSXRay.captureAWSv3Client(s3 as any);
```

Also, tests will not pass because I've removed the transpiled JS from source control, I will add the necessary lifecycle hooks to get tests & publishing working later. Just wanted eyes on the code ASAP.


willarmiros:
@trivikr @AllanZhengYP please review

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -62,110 +79,115 @@ function addFlags(http: HttpResponse, subsegment: Subsegment, err?: SdkError): v
...
-      if (!segment) return next(args);
-
-      const parent = (segment instanceof Subsegment
-        ? segment.segment
-        : segment);
-
-      args.request.headers['X-Amzn-Trace-Id'] = stringify(
+const getXRayPlugin = (config: RegionResolvedConfig, manualSegment?: SegmentLike): Pluggable<any, any> => {
+  const XRayPluginImpl: Pluggable<any, any> = {
+    applyToStack: (stack) => {
```
trivikr, line 92: The variable `XRayPluginImpl` is not updated before returning, so this code can be simplified by removing it.
```suggestion
const getXRayPlugin = (config: RegionResolvedConfig, manualSegment?: SegmentLike): Pluggable<any, any> => ({
  applyToStack: (stack) => {
```

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -62,110 +79,115 @@ function addFlags(http: HttpResponse, subsegment: Subsegment, err?: SdkError): v
...
+      stack.add(
+        (next: any, context: any) => async (args: any) => {
```
trivikr, line 94: Move this function outside Plugin, say in a variable name `xRayMiddleware`?
This way, the plugin and the middleware would be logically separate.

```js
const xRayMiddleware = (next: any, context: any) => async (args: any) => {
  // ...
}

const getXRayPlugin = (config: RegionResolvedConfig, manualSegment?: SegmentLike): Pluggable<any, any> => ({
  applyToStack: (stack) => {
    stack.add(xRayMiddleware, xRayMiddlewareOptions);
  }
});
```

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -18,21 +18,28 @@ var contextUtils = require('../context_utils');
 
 var logger = require('../logger');
```
trivikr, line 19: Use let/const instead of var, as this is a TypeScript file.
```suggestion
const logger = require('../logger');
```

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -18,21 +18,28 @@ var contextUtils = require('../context_utils');
...
 
 async function buildAttributesFromMetadata(
-  client: any,
-  command: any,
-  metadata: ResponseMetadata,
+  service: string,
+  operation: string,
+  region: string,
+  res: any,
 ): Promise<[ServiceSegment, HttpResponse]> {
```
trivikr, line 41: Use arrow function?

```
const buildAttributesFromMetadata = async (
  service: string,
  operation: string,
  region: string,
  res: any,
): Promise<[ServiceSegment, HttpResponse]> => {
```

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -62,110 +79,115 @@ function addFlags(http: HttpResponse, subsegment: Subsegment, err?: SdkError): v
...
-      args.request.headers['X-Amzn-Trace-Id'] = stringify(
+const getXRayPlugin = (config: RegionResolvedConfig, manualSegment?: SegmentLike): Pluggable<any, any> => {
+  const XRayPluginImpl: Pluggable<any, any> = {
+    applyToStack: (stack) => {
+      stack.add(
+        (next: any, context: any) => async (args: any) => {
+          const segment = contextUtils.isAutomaticMode() ? contextUtils.resolveSegment() : manualSegment;
+          const {clientName, commandName} = context;
+          const operation = commandName.slice(0, -7); // Strip trailing "command" string
+          const service = clientName.slice(0, -6);    // Strip trailing "client" string
```
trivikr, line 98:
```suggestion
          const operation = commandName.slice(0, -7); // Strip trailing "Command" string
          const service = clientName.slice(0, -6);    // Strip trailing "Client" string
```

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -62,110 +79,115 @@ function addFlags(http: HttpResponse, subsegment: Subsegment, err?: SdkError): v
...
-
-      args.request.headers['X-Amzn-Trace-Id'] = stringify(
+const getXRayPlugin = (config: RegionResolvedConfig, manualSegment?: SegmentLike): Pluggable<any, any> => {
+  const XRayPluginImpl: Pluggable<any, any> = {
+    applyToStack: (stack) => {
+      stack.add(
+        (next: any, context: any) => async (args: any) => {
+          const segment = contextUtils.isAutomaticMode() ? contextUtils.resolveSegment() : manualSegment;
+          const {clientName, commandName} = context;
+          const operation = commandName.slice(0, -7); // Strip trailing "command" string
```
trivikr, line 97: The commandName is going to start with caps here. You may want to lowerCase first letter, as operation is suffixed to client on line 101

Example commandName for ListTableCommand: https://github.com/aws/aws-sdk-js-v3/blob/2165d4a5c2aec2945900bbe8e79d497da8a80cfa/clients/client-dynamodb/commands/ListTablesCommand.ts#L70

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -62,110 +79,115 @@ function addFlags(http: HttpResponse, subsegment: Subsegment, err?: SdkError): v
...
+            }
+            return next(args);
+          }
+  
+          const subsegment: Subsegment = segment.addNewSubsegment(service);
+          subsegment.addAttribute('namespace', 'aws');
+  
+          const parent = (segment instanceof Subsegment
+            ? segment.segment
+            : segment);
```
trivikr, line 118: AWS SDK for JavaScript team recommends using [prettier](https://prettier.io/) for formatting. 
```suggestion
          const parent = segment instanceof Subsegment ? segment.segment : segment;
```

Your package.json already has prettier, but it wasn't run on this file.
https://github.com/aws/aws-xray-sdk-node/blob/8fa31d45c6276912f36de779f376c7cddb08446e/package.json#L39

I recommend adding a [pre-commit](https://prettier.io/docs/en/precommit.html) hook to run prettier. It can be done as a different PR.

packages/core/lib/patchers/aws3_p.ts
willarmiros: Done, I had to make a `getXRayMiddleware` function so that I could pass in the closure variables.

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -62,110 +82,113 @@ function addFlags(http: HttpResponse, subsegment: Subsegment, err?: SdkError): v
...
+  let res;
+  try {
+    res = await next(args);
+    if (!res) throw new Error('Failed to get response from instrumented AWS Client.');
+
+    // TODO: Figure out how to get region string. Should look like 'await options.region()'
+    const [aws, http] = await buildAttributesFromMetadata(
+      service,
+      operation,
+      await config.region(),
```
willarmiros, line 134: Is this an appropriate way to get region, from the client's `RegionResolvedConfig`?

trivikr: Yes. Here is an example of how it's done in AWS SDK for JavaScript (v3) https://github.com/aws/aws-sdk-js-v3/blob/2165d4a5c2aec2945900bbe8e79d497da8a80cfa/packages/config-resolver/src/EndpointsConfig.ts#L47-L49

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -62,110 +82,110 @@ function addFlags(http: HttpResponse, subsegment: Subsegment, err?: SdkError): v
...
+    },
+    ';',
+  );
 
+  let res;
+  try {
+    res = await next(args);
+    if (!res) throw new Error('Failed to get response from instrumented AWS Client.');
+
+    // TODO: Figure out how to get region string. Should look like 'await options.region()'
```
trivikr, line 130: remove todo, since region is passed?
```suggestion
```

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -62,110 +82,110 @@ function addFlags(http: HttpResponse, subsegment: Subsegment, err?: SdkError): v
...
-      return next(args);
-    },
-    {
-      step: 'build',
-    },
-  );
 
+    const errObj = { message: err.message, name: err.name, stack: err.stack || new Error().stack };
+    subsegment.close(errObj, true);
+    throw err;
```
trivikr, line 160: Is throwing an Object instead of Error expected in Xray SDK?

willarmiros: What do you mean? We construct the `errObj` to record it in the X-Ray subsegment, so it conforms to our data model. But after recording it we just throw the original caught error.

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -14,25 +17,32 @@ import { stringify } from 'querystring';
...
+};
 
-  let operation: string = command.constructor.name.slice(0, -7);
+const buildAttributesFromMetadata = async (
+  service: string,
+  operation: string,
+  region: string,
+  res: any,
+): Promise<[ServiceSegment, HttpResponse]> => {
+  const { extendedRequestId, requestId, httpStatusCode: statusCode, attempts } = res.output?.$metadata || res.$metadata;
```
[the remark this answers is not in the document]
willarmiros: Actually that parameter can be either a `response` object or an `error` object, and the `error` object has `$metadata` populated on it directly. However, I agree that this is a bit of an anti-pattern design so I'll make separate parameters for them.

packages/core/lib/patchers/aws3_p.ts
trivikr: My bad, I mistakenly read as errObj being thrown. There's no issue here.

trivikr (approved):

PR #417, opened 2021-05-06
Add proper TypeScript build logic

willarmiros:
*Issue #, if available:*
#411

*Description of changes:*
These changes allow the repo to conform to TypeScript best practices of not committing transpiled code to source control as we were doing before. Now, we have a `compile` target on the core package which should be run before testing or publishing the package. It transpiles the TypeScript in the source package, and copies the transpiled JavaScript and all other non-TS source files to a `dist` directory, which will ultimately be tested or vended to customers.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -40,6 +40,20 @@ jobs:
           path: $HOME/.npm
           key: ${{ matrix.os }}-${{ hashFiles('package-lock.json') }}
 
+      - name: Install rsync on Windows
+        if: matrix.os == 'windows-latest'
+        run: |
+          ls -al C:/Program Files/Git/usr/bin/
+          echo "performing curl!"
+          curl -sLO http://repo.msys2.org/msys/x86_64/rsync-3.2.3-1-x86_64.pkg.tar.zst
+          echo "decompressing!"
+          ls -al
+          tar -xf rsync-3.2.3-1-x86_64.pkg.tar.zst
+          echo "copying!"
+          ls -al
+          cp rsync.exe C:/Program Files/Git/usr/bin/
+
+
       - name: Execute tests with Lerna
         if: '!matrix.coverage'
         run: |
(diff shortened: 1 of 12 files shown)


.github/workflows/pr-build.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches: [master]
   pull_request:
-    branches: [master]
+    branches: [master, aws-v3-support]
```
anuraaga, line 7: @willarmiros Can you switch this PR to draft?

willarmiros: Ah good call, sorry for the spam

willarmiros:
@anuraaga this should now *actually* be ready for review if you wanna take a look :)

.github/workflows/pr-build.yml
anuraaga, line 7:
```suggestion
    branches: [master]
```

anuraaga (commented):
> not committing transpiled code to source control 

Will you remove the current `.d.ts` files in another PR?

willarmiros:
> Will you remove the current .d.ts files in another PR?

@anuraaga I wasn't sure whether or not I should. We have manually created `.d.ts` files and committed them to source control in all of our packages (not just core). I know the existing declaration files are correct, and was suspect of the generated ones. For instance, in the generated `aws-xray.d.ts`, where we're supposed to declare all of the top-level APIs of the SDK, this was all it contained:

```ts
export namespace middleware {
    const IncomingRequestData: typeof import("./middleware/incoming_request_data");
}
```

I'm sure we can make a series of changes to our source to make the generated declaration files work, but given that TS customers have been happy with the existing ones I'm not sure I'll prioritize it along with these changes.

willarmiros:
I'm going to keep CI enabled on the `aws-v3-support` and go ahead and merge this so I can incorporate these changes with #416. I'll disable CI and we can re-review when I make the PR for merging `aws-v3-support` into master.

anuraaga:
Oh ok if those are manual b declarations than sounds good didn't quite understand what's going on.

PR #421, opened 2021-05-11
Re-add AWS V3 Instrumentation

willarmiros:
*Issue #, if available:*
#294, #411

*Description of changes:*
This PR officially re-commits AWS SDK V3 instrumentation to the main branch of the repo. To summarize the changes, we have:
1. Redesigned the instrumentation to entirely use the AWS SDK V3 middleware tools instead of monkey-patching
2. Changed the publishing mechanism of the `core` package to exclude TS files and only include type declarations and transpiled JS code, as is best practice
3. Updated the types used in our type declaration and instrumentation modules so that we minimize added dependencies while still able to keep the same TypeScript experience AWS SDK V3 customers expect

Generally, the changes in this PR have already been reviewed in #386, #416, and #417. The only additional changes are those mentioned in point 3.


packages/core/README.md
```diff
@@ -465,6 +465,8 @@ function sendRequest(host, cb) {
 
 ### Capture all outgoing AWS requests
 
+This is only available for AWS SDK v2 due to the service-oriented architecture of AWS SDK v3.
```
lupengamzn, line 468: Probably good to add a link to the announcement about the difference between v2 and v3

lupengamzn (commented):
One minor comment, otherwise LGTM

packages/core/README.md
willarmiros: Good idea, just added

lupengamzn (approved):

Issue #427, opened 2021-05-13
Cannot find module 'aws-xray-sdk-core/lib/middleware/sampling/service_connector' from 'node_modules/aws-xray-sdk-express/lib/express_mw.js'


ls-jacob-lawrence:
`Cannot find module 'aws-xray-sdk-core/lib/middleware/sampling/service_connector' from 'node_modules/aws-xray-sdk-express/lib/express_mw.js'`

Version 3.3.2 introduces this bug due to 84e65c3f2c9119a40f88c5e6ea1d9070a0a8ca7f

The absolute path is missing the /dist/ for consumers

ChungminPark:
I had an issue derived from the new /dist/ path behind the /lib/ because I directed accessed to `aws-xray-sdk-core/lib/context_utils` on my project. so I got `Cannot find module 'aws-xray-sdk-core/lib/context_utils` error. I solved this issue by fixing `aws-xray-sdk-core` version to downgraded one("aws-xray-sdk-core": "^3.1.0" -> "aws-xray-sdk-core": "3.3.1" on package.json).

estiller:
This happened to our team as well. All newly built Docker images stopped loading. We worked around it by setting the version to exactly "3.3.1" on package.json.

padzikm:
We have the same problem with error Runtime.ImportModuleError: Error: Cannot find module 'aws-xray-sdk-core/lib/middleware/sampling/service_connector' - version 3.2.0 all packages aws-xray-sdk-*

willarmiros:
Hello,

Thanks for raising this issue. We'll get a patch out ASAP to address it.

willarmiros:
@ChungminPark we intend on keeping the new structure with `dist` in the path. All of the public APIs in the `context_utils` module are exposed via the top-level `AWSXRay` object, e.g. `AWSXRay.setSegment()`, and I would recommend referencing them like that so you aren't dependent on the directory structure.

GregTurner:
Can we get a npm publish please?

willarmiros:
@GregTurner preparing the release right now.

willarmiros:
This issue is fixed in v3.3.3, available now on NPM. Please let us know if there are any further issues. Apologies again for the trouble.

ChungminPark:
@willarmiros Thanks a lot!

PR #428, opened 2021-05-13
updated non-dist paths

willarmiros:
*Issue #, if available:*
#427

*Description of changes:*
Updated references to core package that used absolute path without `dist`.

### DIFF
--- a/packages/express/test/unit/express_mw.test.js
+++ b/packages/express/test/unit/express_mw.test.js
@@ -5,8 +5,8 @@ var sinon = require('sinon');
 var sinonChai = require('sinon-chai');
 
 var expressMW = require('../../lib/express_mw');
-var SegmentEmitter = require('../../../core/lib/segment_emitter.js');
-var ServiceConnector = require('../../../core/lib/middleware/sampling/service_connector.js');
+var SegmentEmitter = require('../../../core/dist/lib/segment_emitter.js');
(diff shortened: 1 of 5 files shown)


packages/express/test/unit/express_mw.test.js
```diff
@@ -5,8 +5,8 @@ var sinon = require('sinon');
 var sinonChai = require('sinon-chai');
 
 var expressMW = require('../../lib/express_mw');
-var SegmentEmitter = require('../../../core/lib/segment_emitter.js');
-var ServiceConnector = require('../../../core/lib/middleware/sampling/service_connector.js');
+var SegmentEmitter = require('../../../core/dist/lib/segment_emitter.js');
```
lupengamzn, line 8: Is the `dist` repo going to be existed in the final assets that are packaged?

willarmiros: Yep, the change in 3.3.2 was to repackage all of the content from `lib` into a `dist` directory, which is what's ultimately tested against and vended to customers.

willarmiros: These unit tests were passing before because when you clone the source repo and build the SDK, you have access to BOTH the `dist` and `lib` directories. But once it is vended to customers, only the `dist` is available.

lupengamzn: Thanks for the clarification as I was thinking why unit tests didn't catch this

lupengamzn (approved):

Issue #401, opened 2021-04-04
Wrong http.request.url reported when host header is set


danilobuerger:
Sometimes it is useful to connect to a url and have a different host header:

```
   The "Host" header field in a request provides the host and port
   information from the target URI, enabling the origin server to
   distinguish among resources while servicing requests for multiple
   host names on a single IP address.
```
https://tools.ietf.org/html/rfc7230#section-5.4

When this is the case, the reported `http.request.url` contains the host header instead of the host from the connecting url.

For example:

```
fetch("http://url.example/path", { headers: { host: "host.example" }})
```

would be reported as `http://host.example/path` instead of `http://url.example/path`.

willarmiros:
Hi @danilobuerger,

That's a fair calllout, we do parse the host header for the *name* of the subsegment [here](https://github.com/aws/aws-xray-sdk-node/blob/811d9582103cafcf948dade3efaef1b92aa211bd/packages/core/lib/patchers/http_p.js#L98), and for the `http.request.url` field as you said [here](https://github.com/aws/aws-xray-sdk-node/blob/811d9582103cafcf948dade3efaef1b92aa211bd/packages/core/lib/segments/attributes/remote_request_data.js#L17). I would be open to a PR to prefer parsing the hostname from the URL instead.

danilobuerger:
@willarmiros Thanks for your response. Sadly I don't have the time to submit a proper PR at this time. However, here is my patch file if somebody is willing to submit it as a PR or runs into the same issue:

```diff
diff --git a/node_modules/aws-xray-sdk-core/lib/patchers/http_p.js b/node_modules/aws-xray-sdk-core/lib/patchers/http_p.js
index 7743ecd..6ea5fc7 100644
--- a/node_modules/aws-xray-sdk-core/lib/patchers/http_p.js
+++ b/node_modules/aws-xray-sdk-core/lib/patchers/http_p.js
@@ -135,7 +135,7 @@ function enableCapture(module, downstreamXRayEnabled, subsegmentCallback) {
       } else {
         var madeItToDownstream = (e.code !== 'ECONNREFUSED');
 
-        subsegment.addRemoteRequestData(this, null, madeItToDownstream && downstreamXRayEnabled);
+        subsegment.addRemoteRequestData(this, options, null, madeItToDownstream && downstreamXRayEnabled);
         subsegment.close(e);
       }
 
[...]

PR #430, opened 2021-05-17
parse hostname from options instead of headers

willarmiros:
*Issue #, if available:*
#401

*Description of changes:*
Adds optional parameter for `options` object, from which we can read the `hostname` field instead of the headers. In most cases, this will be exactly the same string, but as pointed out in #401 sometimes the `host` header has a different host from the actual target's hostname.

### DIFF
--- a/packages/core/lib/segments/attributes/remote_request_data.js
+++ b/packages/core/lib/segments/attributes/remote_request_data.js
@@ -7,15 +7,17 @@ var { stripQueryStringFromPath } = require('../../utils');
  * @param {http.ClientRequest|https.ClientRequest} req - The request object from the HTTP/HTTPS call.
  * @param {http.IncomingMessage|https.IncomingMessage} res - The response object from the HTTP/HTTPS call.
  * @param {boolean} downstreamXRayEnabled - when true, adds a "traced": true hint to generated subsegments such that the AWS X-Ray service expects a corresponding segment from the downstream service.
+ * @param {object} [options] - Optional HTTP Request options, which will take precedence over header data to populate request metadata.
  */
 
-function RemoteRequestData(req, res, downstreamXRayEnabled) {
-  this.init(req, res, downstreamXRayEnabled);
+function RemoteRequestData(req, res, downstreamXRayEnabled, options) {
+  this.init(req, res, downstreamXRayEnabled, options);
 }
 
-RemoteRequestData.prototype.init = function init(req, res, downstreamXRayEnabled) {
+RemoteRequestData.prototype.init = function init(req, res, downstreamXRayEnabled, options) {
+  const useOptionsHost = options && options.hostname;
   this.request = {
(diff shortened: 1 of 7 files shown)


packages/core/lib/segments/attributes/remote_request_data.js
```diff
@@ -7,15 +7,17 @@ var { stripQueryStringFromPath } = require('../../utils');
...
 
-function RemoteRequestData(req, res, downstreamXRayEnabled) {
-  this.init(req, res, downstreamXRayEnabled);
+function RemoteRequestData(req, res, downstreamXRayEnabled, options) {
+  this.init(req, res, downstreamXRayEnabled, options);
 }
 
-RemoteRequestData.prototype.init = function init(req, res, downstreamXRayEnabled) {
+RemoteRequestData.prototype.init = function init(req, res, downstreamXRayEnabled, options) {
+  const useOptionsHost = options && options.hostname;
```
anuraaga, line 18: Isn't the original issue wanting to read from the request url, not a fixed option?

willarmiros: The original issue said it was problematic to read from the header, and instead preferred reading from the `options` object. But I guess reading from the `request` host would work just as well and not require a signature change.

anuraaga: Yeah let's just change the logic to read the request URL. It seems like a bug we use the host header (should suspect other languages / instrumentations aren't doing so) so just a simple be bugfix

willarmiros: Updated to prefer `request.host` and fall back to the header value. It looks like `request.host` was only [added in Node 12](https://nodejs.org/api/http.html#http_request_host), so I guess that's why it was this way before.

packages/core/test/unit/segments/attributes/remote_request_data.test.js
```diff
@@ -50,5 +50,14 @@ describe('RemoteRequestData', function() {
         ''
       );
     });
+    it('should use the host from the request object over headers', () => {
+      const requestWithHost = Object.assign(request, { host: 'different-site.com' });
```
anuraaga, line 54: Hmm - could cleanup in another PR, `getHeader() { return 'host.com' }` seems like very suspicious code :) But I guess that's what makes sure we're overwriting the header.

By the way, if not in a unit test is it possible to verify the behavior with `fetch` which was the example in the filed issue? I assume it's `node-fetch`. Don't see a reason it wouldn't work though

https://github.com/node-fetch/node-fetch/blob/master/src/request.js#L211

anuraaga (approved):

packages/core/test/unit/segments/attributes/remote_request_data.test.js
willarmiros: I updated the `getHeader` to be less sus :) 

As for testing `node-fetch`, we could and probably should, but that would require a whole lot more stubbing than is probably in scope for this change since we're only testing the underlying Node HTTP client today.

PR #432, opened 2021-05-20
Added CI smoke test workflow

willarmiros:
*Issue #, if available:*
Prevent another #427

*Description of changes:*
Introduces a new workflow to be ran on all PRs, which runs our simple smoke test against the locally-packaged bundles that are identical to what the end-user would receive. We also remove all the original source code to ensure there are no conflicts with running the tests from within the repo.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -17,6 +17,8 @@ jobs:
         with:
           node-version: '14.x'
 
+      # Use npm pack to bundle individual packages into their final distribution tarball form
+      # Then aggregate all of those tarballs in the full_sdk package, to be used by smoke test
       - name: Pack SDK packages
         id: package
         run: |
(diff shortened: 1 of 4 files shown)


willarmiros:
As we can see, the [smoke test failed](https://github.com/aws/aws-xray-sdk-node/runs/2634173188) on [a commit](https://github.com/aws/aws-xray-sdk-node/pull/432/commits/645d21154eb1ed9e02352685c7dd4fa4832e76bc) that reproduced the bad import issue, even when the unit tests passed.

When we removed the bad commit in `d865e4d`, all tests passed again.

.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,46 @@
+name: Node.js Continuous Integration Smoke Test
+
+on:
+  pull_request:
+    branches: [master]
```
anuraaga, line 5: I was about to wonder about unit tests but see the file called `pr-build.yml` runs on main branch push :P

anuraaga (approved):

PR #434, opened 2021-05-25
Fix crash when http/https libraries use getters

### LINKED ISSUE #433: captureHTTPsGlobal fails when bundled with esbuild
When using AWS CDK to deploy Lambda functions with [aws-lambda-nodejs](https://docs.aws.amazon.com/cdk/api/latest/docs/aws-lambda-nodejs-readme.html), trying to call:
```
xray.captureHTTPsGlobal(http, false)
```
from aws-xray-sdk-node>=3.3.2 causes the function to fail with:
```
Cannot set property request of #<Object> which has only a getter
```
This is caused by [esbuild](https://github.com/evanw/esbuild) which aws-lambda-nodejs uses to bundle the code. For details, see: https://github.com/evanw/esbuild/issues/587

Now while this is really a bug in the bundler, it would be easy to fix this in aws-xray-sdk-node, and I'd be happy to make a PR.

christophgysin:
*Issue: #433*

*Description of changes:*

This works around an issue when using esbuild to bundle (e.g. using aws-lambda-nodejs)

The fix is adapted from:
https://github.com/follow-redirects/follow-redirects/pull/146

fixes #433

### DIFF
--- a/packages/core/lib/patchers/http_p.js
+++ b/packages/core/lib/patchers/http_p.js
@@ -196,14 +196,19 @@ function enableCapture(module, downstreamXRayEnabled, subsegmentCallback) {
   }
 
   module.__request = module.request;
-  module.request = function captureHTTPsRequest(...args) {
+  function captureHTTPsRequest(...args) {
     return captureOutgoingHTTPs(module.__request, ...args);
-  };
+  }
 
   module.__get = module.get;
-  module.get = function captureHTTPsGet(...args) {
+  function captureHTTPsGet(...args) {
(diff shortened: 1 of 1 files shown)


srprash (approved):
LGTM. @willarmiros Can you take a quick look as well?

willarmiros (approved):
LGTM

cnuss:
@willarmiros can we get a release soon that includes this? thanks!

willarmiros:
Sorry for the delay on this @cnuss - should have a release out soon!

Issue #329, opened 2020-08-28
Only one eslintrc


anuraaga:
We currently have an eslintrc for each package, but we want a consistent style in the repo. Instead of copying it everywhere, we should have just one.

PR #436, opened 2021-05-28
Added enforced linting

willarmiros:
*Issue #, if available:*
#329 

*Description of changes:*
Added a top-level `.eslintrc.json` which contains all config, and just extended from it in all subpackages. We use the `@typescript-eslint/recommended` and `import/recommended` base configurations, which don't end up being too opinionated, but is in line with what the AWS SDK uses. I considered adding Prettier, and am still not opposed, but neither the AWS SDK nor OpenTelemetry JS uses it (see https://github.com/open-telemetry/opentelemetry-js/pull/2204 for more on the recent removal).

It turns out I can't really add a linting rule to prevent an import like the one that caused #427 because it's a perfectly valid import, but such issues should be caught by the smoke tests now anyway and the repo was long overdue for linting.


willarmiros:
Looks like there's some weirdness with the `hapi` package linting. Will address before merging.

.eslintrc.json
```diff
@@ -0,0 +1,52 @@
...
+            "error",
+            "always"
+        ],
+        "eol-last": [
+            "error",
+            "always"
+        ],
+        "no-trailing-spaces": "error",
+        
+        /** Turn off strict enforcement */
```
anuraaga, line 38:
```suggestion
        /** Turn off enforcement */
```

I think off turns off completely, not just make it "not strict

packages/core/lib/middleware/sampling/rule_cache.js
```diff
@@ -55,7 +55,7 @@ var RuleCache = {
       if(v !== 0) return v;
       if(a.getName() > b.getName())
         return 1;
```
anuraaga, line 57: I'd expect our lint to enforce braces around blocks

anuraaga (approved):

packages/core/lib/middleware/sampling/rule_cache.js
willarmiros: Good call, will add the `curly` rule

Issue #439, opened 2021-06-11
captureAWSv3Client does not support SQS or SSM clients


thesuavehog:
The `captureAWSv3Client` documentation states:
```javascript
/**
 * Instruments AWS SDK V3 clients with X-Ray via middleware.
 *
 * @param client - AWS SDK V3 client to instrument
 * @param manualSegment - Parent segment or subsegment that is passed in for manual mode users
 * @returns - the client with the X-Ray instrumentation middleware added to its middleware stack
 */
```

However if you try and use [SSMClient](https://github.com/aws/aws-sdk-js-v3/tree/main/clients/client-ssm) or [SQSClient](https://github.com/aws/aws-sdk-js-v3/tree/main/clients/client-sqs) in a TypeScript context you will get an error:
```
Argument of type 'SSMClient' is not assignable to parameter of type 'Client<any, any, any>'.
  The types of 'middlewareStack.concat' are incompatible between these types.
     ... and many other lines...
```
Presumably this applies to any AWS v3 Client which extends
`Client<HandlerOptions, ClientInput extends object, ClientOutput extends MetadataBearer, ResolvedClientConfiguration extends SmithyResolvedConfiguration<HandlerOptions>>`
(in `@aws-sdk/smithy-client/dist/types/client.d.ts`).

```typescript
import { SSMClient } from '@aws-sdk/client-ssm';
import { captureAWSv3Client } from 'aws-xray-sdk';
let ssm: SSMClient = new SSMClient({});
ssm = captureAWSv3Client(ssm);
```

```
Argument of type 'SSMClient' is not assignable to parameter of type 'Client<any, any, any>'.
  The types of 'middlewareStack.concat' are incompatible between these types.
    Type '<InputType extends ServiceInputTypes, OutputType extends ServiceOutputTypes>(from: MiddlewareStack<InputType, OutputType>) => MiddlewareStack<InputType, OutputType>' is not assignable to type '<InputType extends any, OutputType extends any>(from: MiddlewareStack<InputType, OutputType>) => MiddlewareStack<InputType, OutputType>'.
      Types of parameters 'from' and 'from' are incompatible.
        Type 'MiddlewareStack<InputType, OutputType>' is not assignable to type 'MiddlewareStack<InputType, ServiceOutputTypes>'.
          Types of property 'addRelativeTo' are incompatible.
[...]

PR #449, opened 2021-07-12
Moved @aws-sdk/types to peer dependencies to resolve incompatibilities

willarmiros:
*Issue #, if available:*
#439

*Description of changes:*
As title, based off [customer's comment](https://github.com/aws/aws-xray-sdk-node/issues/439#issuecomment-878317558) and AWS SDK recommendation. This way, we will be able to use whatever version of `@aws-sdk/types` is provided by the customer, which should resolve the error reported in #439. Tested on a sample app to verify instrumentation still works.


willarmiros:
Actually, after testing this with verdaccio, I realized this change will cause an issue similar to #411 in that users of this SDK who do *not* use the AWS SDK V3 (and therefore do not have `@aws-sdk/types` package available) will fail to compile. Therefore to maintain functionality for *all* users of this SDK, we must keep the types package as a proper dependency.

m-radzikowski:
I think that with npm v6 and yarn, the non-installed peer dependencies are just logged as a warning during the install. And I see the npm v7 installs them by default now (https://stackoverflow.com/questions/35207380/how-to-install-npm-peer-dependencies-automatically).

I recall AWS SDK or some other libs have peer dependency on some frontend stuff for react etc., that is not needed nor installed when developing for Node.JS.

Let me check how it behaves on my own as well.

willarmiros:
That is interesting that npm v7 installs them by default - sounds like that will solve this long-term but I think many people are still using npm 6 or less. It does in fact work with native JS (as long as you don't call the `captureAWSV3Client` API, which you wouldn't if you don't have an AWS SDK V3 dependency). Where it fails for me is running `tsc` on a project that uses `aws-xray-sdk-core` and does not satisfy the `@aws-sdk/types` dependency.

m-radzikowski:
Ok, I reproduced the same on my own, and I see the problem.

I hoped that this: https://stackoverflow.com/a/63905376/2512304 can be the solution, but a quick test with changing the

```typescript
import { Client } from '@aws-sdk/types';
```

to

```typescript
import type { Client } from '@aws-sdk/types';
```

still results in a `tsc` error when `@aws-sdk/types` is not installed (in the client module that uses `aws-xray-sdk-core`). I either did something wrong or it's not a solution for this problem (again, only did a quick test).

I see that for the AWS SDK v2 you ignore the type completely: https://github.com/willarmiros/aws-xray-sdk-node/blob/peer-types/packages/core/lib/patchers/aws_p.d.ts#L7

While it would be best to have the `captureAWSv3Client()` fully typed, maybe it's better to not use `@aws-sdk/types` at all and do a similar signature here?

```typescript
export declare function captureAWSClient<T>(client: T, manualSegment?: SegmentLike): T
```

As I see it:

- right now, the `client` type is checked only if I match the same `@aws-sdk/types` version as the xray lib uses,
- otherwise, I have to disable type checking on my own (with `client as any`).

Since you won't be releasing a new xray lib version every time there is a new AWS SDK v3 release, keeping `@aws-sdk/types` in sync would require using the old AWS SDK v3 lib version (that xray lib uses) and never update. In practice, I guess everyone will have to use `as any`. So from my perspective, it's far better to release the type restriction on the xray lib itself, getting rid of `@aws-sdk/types` completely here and making the function signature similar to the one for AWS SDK v2.

Well, it's my opinion and I can be wrong here, so please let me know what you think. It just hurts me when I have to do `captureAWSv3Client(client as any)` in my code 😄

thesuavehog:
@willarmiros Can there be a decision made about how to deal with this issue? Either kill this PR so it doesn't appear in searches as a possible solution, or make a PR that addresses the problem by removing types. I have trained up 5 developers on using XRay now and every single one of them would have run into this issue and been confused if I hadn't built my own wrapper library around XRay to specifically `as any` (and then I have to explain to them why they have to use my library and ignore all the online documentation about how to wrap AWS SDK clients in XRay).

It seems like a pretty major issue that capture simply _does not work_ per all the documentation about how to use it for a (growing) number AWS SDK v3 clients.

package.json
```diff
@@ -4,10 +4,11 @@
...
   "devDependencies": {
-    "@aws-sdk/config-resolver": "^3.4.1",
-    "@aws-sdk/middleware-stack": "^3.3.0",
-    "@aws-sdk/node-config-provider": "^3.4.1",
-    "@aws-sdk/smithy-client": "^3.4.1",
+    "@aws-sdk/config-resolver": "^3.18.0",
+    "@aws-sdk/middleware-stack": "^3.18.0",
+    "@aws-sdk/node-config-provider": "^3.18.0",
+    "@aws-sdk/smithy-client": "^3.18.0",
+    "@aws-sdk/types": "^3.18.0",
```
NathanielRN, line 11: Did you mean `^3.0.0` to maximize compatibility?

```suggestion
    "@aws-sdk/types": "^3.0.0",
```

NathanielRN (commented):
Just some passing comments.

NathanielRN:
@thesuavehog sorry for your frustration with this issue. I'll add it to the agenda for our team to discuss next week.

In the meantime have you considered using [the AWS Distro for OpenTelemetry (ADOT) - JavaScript](https://aws-otel.github.io/docs/getting-started/javascript-sdk) solution?

OTel JS gets a lot of support from many devs and [we announced GA support for it](https://aws.amazon.com/about-aws/whats-new/2021/09/aws-availability-distro-opentelemetry-tracing-support/) last year. All components are upstream and ready for you to use to trace with OTel JS and AWS X-Ray 🙂 We've been recommending users check that out so maybe you'll find excellent instrumentation there with minimal changes needed to get it working?

Let me know if you have any questions!

willarmiros:
Superseded by #575.

Issue #452, opened 2021-07-16
MySQL capture failing for mysql2 pool queries with promises


simoncollins:
We're looking to switch our existing code that uses `mysql2` to use X-Ray instrumentation. We have hit issues with using either the parent `mysql2` module or `mysql2/promise`.

Our code is essentially one of these two scenarios (both fail):

## With parent mysql2 module and `.promise()`
```js
const rawMysql = require('mysql2');
const captureMySQL = require('aws-xray-sdk-mysql');

(async function() {
  const mysql = captureMySQL(rawMysql);
  
  const pool = mysql.createPool({
      // ... connection setting ...
  }).promise();
  
  await pool.query("SELECT ...");
})();
```

This fails with:

```
Error: You have tried to call .then(), .catch(), or invoked await on the result of query that is not a promise, which is a programming error
```

at `mysql2/lib/commands/query.js:41:11` because the mysql2 Query class has a stubbed `then` method that throws this error.

In `captureOperation` in `mysql_p.js` the code tries to detect whether the command object is promise or event emitter based and this stubbed method confuses it into thinking it's actually a promise.

I'm not sure what can be done about this since there's no reliably better way to detect a promise.

## With the `mysql2/promise` module
```js
const rawMysql = require('mysql2/promise');
const captureMySQL = require('aws-xray-sdk-mysql');

(async function() {
  const mysql = captureMySQL(rawMysql);
  
  const pool = mysql.createPool({
      // ... connection setting ...
  });
  
  await pool.query("SELECT ...");
})();
```

This fails for a different reason in `captureOperation` where it tries to grab the connection config object using:
 ```js
var config = this.config.connectionConfig || this.config
```

This is because the `mysql2/promise` module returns a `PromisePool` instance which delegates to the actual pool object holding the connection config object and there is no `config` property on the base object.

This does seem like something that could be addressed in the operation patching code by detecting and handling this scenario. Something like:
```js
var config = this.pool ?
        this.pool.config.connectionConfig :
        (this.config.connectionConfig || this.config);
```

Has anyone else experienced this issue? I noticed that there is some support for mysql2 already so I'm wary that I'm doing something wrong having not used either library before. Alternatively it might be that it's our use of the pool object that is not covered by existing scenarios?

simoncollins:
For the moment this workaround of patching the pool object works for me:
```js
function patchPool(promisePool) {
    return {
        ...promisePool,
        config: promisePool.pool.config
    }
}

[...]

PR #501, opened 2022-04-26
fix: avoid fake .then() method on mysql2 Query class

mickdekkers:
*Issue #, if available:* #452

*Description of changes:*

The [mysql2 package](https://github.com/sidorares/node-mysql2) has a `Query` class with [a `.then()` method which throws an error when called](https://github.com/sidorares/node-mysql2/blob/dbb344e89a1cc8bb457b24e67b07cdb3013fe844/lib/commands/query.js#L38-L44).
This confuses the current patching code, which sees this object as a `Promise` and calls `.then()` on it.

As a result, code like this currently breaks with an error:
```js
const captureMySQL = require('aws-xray-sdk-mysql');
const {createConnection} = captureMySQL(require('mysql2'));

const connection = createConnection(someConfig);
const queryStream = connection
        .query(someSql) // <-- returns Query class, error is thrown here
        .stream({highWaterMark: 10});

// (Attempt to) do stuff with queryStream
queryStream.pipe(someFileOnDisk);
```

This PR fixes this unfortunate interaction by performing an additional check to determine whether a returned value is actually a `Promise`.

Beyond the new unit test (which fails without the change and passes with it), I can confirm this change has fixed the error in our application.

### DIFF
--- a/packages/mysql/lib/mysql_p.js
+++ b/packages/mysql/lib/mysql_p.js
@@ -33,13 +33,26 @@ module.exports = function captureMySQL(mysql) {
   return mysql;
 };
 
+function isPromise(maybePromise) {
+  if (maybePromise != null && maybePromise.then instanceof Function) {
+    // mysql2 has a `Query` class with a `then` method which always throws an error when called.
+    // We want to avoid calling this, so we need to check for more than just the presence of a `then` method.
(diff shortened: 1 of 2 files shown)


willarmiros (approved):
Thanks for this fix!

Issue #450, opened 2021-07-13
TypeError "aws_xray_sdk_mysql_1.captureMySQL is not a function"


richardcornelissen:
We've recently updated `aws-xray-sdk-mysql` from version `3.2.0` to `3.3.3`, but experienced issues with our Typescript projects.

Due to the changed Typescript definitions, we had to adjust the import from
```typescript
import captureMySQL from 'aws-xray-sdk-mysql';
```

to

```typescript
import {captureMySQL} from 'aws-xray-sdk-mysql';
```

However, this resulted in errors as such:
```log
{
    "errorType": "TypeError",
    "errorMessage": "aws_xray_sdk_mysql_1.captureMySQL is not a function",
    "stack": [
        "TypeError: aws_xray_sdk_mysql_1.captureMySQL is not a function",
        "    at Object.exports.setup (/var/task/src/api.js:2:1851104)",
        "    at Object.<anonymous> (/var/task/src/api.js:2:1848725)",
        "    at __webpack_require__ (/var/task/src/api.js:2:295)",
        "    at Object.<anonymous> (/var/task/src/api.js:2:1847607)",
        "    at __webpack_require__ (/var/task/src/api.js:2:295)",
        "    at /var/task/src/api.js:2:1594",
        "    at Object.<anonymous> (/var/task/src/api.js:2:1641)",
        "    at Module._compile (internal/modules/cjs/loader.js:999:30)",
        "    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)",
        "    at Module.load (internal/modules/cjs/loader.js:863:32)"
    ]
}
```

**Workaround**

As a temporary workaround, replace

```typescript
import {captureMySQL} from 'aws-xray-sdk-mysql';
```

With

```typescript
// eslint-disable-next-line @typescript-eslint/no-var-requires
const captureMySQL = require('aws-xray-sdk-mysql');
```

willarmiros:
Hi @richardcornelissen,

I wasn't able to reproduce this with a normal TypeScript app importing `captureMySQL` normally. I noticed you're using webpack, so I think it's likely related to some of the magic that does under the hood. Thank you for providing the workaround, there are some known issues with this SDK and webpack so I will label this as a bug within that category for now.

beanaroo:
Can confirm that we are getting the same issue:
```
capturePostgres is not a function
```

Irvel:
[...]

PR #504, opened 2022-05-04
capturePostgres is the default function

hckhanh:
It must be equivalent to require import:

```js
var capturePostgres = require('aws-xray-sdk-postgres');
```

I get this error when try to follow the type definition of this package with TypeScript

```ts
import { capturePostgres } from "aws-xray-sdk-postgres";
```

Then I get this error:

```sh
TypeError: (0 , external_aws_xray_sdk_postgres_namespaceObject.capturePostgres) is not a function
```

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/postgres/lib/postgres_p.d.ts
+++ b/packages/postgres/lib/postgres_p.d.ts
@@ -1,7 +1,7 @@
 import * as AWSXRay from 'aws-xray-sdk-core';
 import * as PG from 'pg';
 
-export function capturePostgres(pg: typeof PG): capturePostgres.PatchedPostgres;
+export default function capturePostgres(pg: typeof PG): capturePostgres.PatchedPostgres;
 
 declare namespace capturePostgres {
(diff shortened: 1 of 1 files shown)


willarmiros:
Hi @hckhanh it seems this change is causing linting issues could you address those?

hckhanh:
Hi @willarmiros, the linting error is `No named exports found in module './postgres_p'`
I have to change the interface of the implementation to match the ESLint rule. Is that ok?

hckhanh:
Or I simply change the lint rule to make it success.

willarmiros:
> I have to change the interface of the implementation to match the ESLint rule. Is that ok?

We can't make any backwards incompatible/breaking changes. I would recommend ignoring the lint rule in code and adding a comment explaining why we suppress it.

mxdvl:
Any chance we could get this over the line so that we can close #450?

PR #512, opened 2022-05-27
Remove support for Node 10.x

NathanielRN:
*Issue #, if available:* N/A

*Description of changes:*

We recently noticed this error on the AWS Java X-Ray SDK [Continuous build workflows](https://github.com/aws/aws-xray-sdk-node/actions/runs/2393563159)

```
npx: installed 650 in 39.37s
Object.fromEntries is not a function
```

We have a package called `fromentries` but I don’t see it used anywhere…

We found online that [there is an issue with Node v10](https://stackoverflow.com/a/67551131/7460739) in which this error will show up. We are testing with Node v10 GitHub Actions so our solution is to drop Node v10 support to unblock our CI. [Node 10 is not supported by Node themselves](https://nodejs.org/en/about/releases/) according to their page

This is blocking us from releasing the Node X-Ray SDK.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -17,7 +17,6 @@ jobs:
           - ubuntu-latest
           - windows-latest
         node-version:
-          - 10.x
           - 12.x
           - 14.x
           - 15.x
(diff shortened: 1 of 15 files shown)


willarmiros (approved):
Looks good for removing v10, but let’s be sure to add the latest versions of node in a future PR.

Issue #460, opened 2021-08-24
xray whitelist for additional data capture does not work with AWS Javascript v3 SDK clients


thesuavehog:
Using X-Ray with the AWS Javascript v3 SDK clients prints out debug logging messages that indicate the additional data captured in the v2 clients via the whitelist is not applied to the v3 clients.

Example Log:
```
Call \"DynamoDB.GetItem\" is not whitelisted for additional data capturing. Ignoring.
```
(similar messages appear for other SDK API calls)

This is the default whitelist that is included in the AWS X-Ray SDK:
[aws_xray_whitelist.json.txt](https://github.com/aws/aws-xray-sdk-node/files/7041056/aws_xray_whitelist.json.txt)

I have tried manually setting the whitelist (see below) and leaving it unset (the default behavior).
```javascript
import AWSXRay from 'aws-xray-sdk';
const whitelist: any = require('aws-xray-sdk-core/dist/lib/resources/aws_whitelist.json');
AWSXRay.setAWSWhitelist(whitelist);
```

When I look at the trace in the X-Ray console, I see the following in the Resources tab (for the trace):
```
Retries    | 1
Region     | us-east-1
Operation  | GetItem
Request ID | USTUCFQJA974H461PMB95LLL73VV4KQNSO5AEMVJF66Q9ASUAAJG
```
the annotations, resources and exceptions tabs have no data (for successful requests).

Is there a special way to enabled the extra data capture for v3 SDK clients?

willarmiros:
HI @thesuavehog,

Unfortunately this feature for capturing additional data is not yet available with v3 clients because the AWS SDK v3 uses a different data model for storing request/response attributes. We will keep this open as a feature request for adding that support. One possible alternative for a workaround would be to add hooks for the AWS SDK v3 patcher so that custom logic could be added to capture these attributes.

thesuavehog:
@willarmiros re: `add hooks for the AWS SDK v3 patcher`
Is that something I can do on my project or is that something that has to be done on the SDK level?

thesuavehog:
also, there's a LOT of `debug` output from X-Ray related to the whitelisting which is annoying when trying to parse out my own debug logs. If this is not a supported feature, then X-Ray should not output the debug logs on EVERY call (in my opinion).

If you want to output 1 log item when a client is initially wrapped as an FYI, that's fine, but the current verboseness is annoying =)

willarmiros:
[...]

PR #527, opened 2022-08-16
fix(core): capture additional command data with SDK v3

SergeyPoluektov:
*Issue #, if available:* #460

*Description of changes:*

Additional data is added to a AWS segment via [`Capturer.capture`](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/patchers/call_capturer.js#L53) method. The method uses a [whitelist](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/resources/aws_whitelist.json) for allowed services and operations to be captured. The capturer is used by a [AWS service segment](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/segments/attributes/aws.js#L15) to build a segment with additional request data. The first argument there – the response object from the AWS call, is used to add metadata to a segment with capturer.

The [AWS SDK v3 patcher](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/patchers/aws3_p.ts#L48) passes a response object to a AWS service segment without `data` and `request.params` attributes. These attributes are used by the capturer to obtain additional metadata from the service call.

The `response.data` attribute is used to capture `response_parameters` or `response_descriptors` such as [`ConsumedCapacity`](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/resources/aws_whitelist.json#L41) (see [`Capturer.capture`](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/patchers/call_capturer.js#L71)).
And the `response.request.params` is used to capture `request_parameters` or `request_descriptors` such as [`TableName`](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/resources/aws_whitelist.json#L38) (see [`Capturer.capture`](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/patchers/call_capturer.js#L67))

**This PR adds `data` and `request.params` properties to a first parameter passed to the AWS service segment:**
- the client response output `res.output` is passed to a `data` property of the AWS service segment to capture response parameters and descriptors.
- the command input `command.input` is passed to a `request.params` property of the AWS service segment to capture request parameters and descriptors. Assuming that every `Command` in AWS SDK v3 has `input` property

I used [this example](https://github.com/SergeyPoluektov/lambda-dynamo-example/blob/main/src/app.ts) app (lambda writing to DynamoDB) to test traces. 
**Trace before changes:**
image
image

**Trace after changes:**
image
image

### DIFF
--- a/packages/core/lib/patchers/aws3_p.ts
+++ b/packages/core/lib/patchers/aws3_p.ts
@@ -40,6 +40,7 @@ const buildAttributesFromMetadata = async (
   service: string,
   operation: string,
   region: string,
+  commandInput: any,
   res: any | null,
   error: SdkError | null,
 ): Promise<[ServiceSegment, HttpResponse]> => {
@@ -50,8 +51,10 @@ const buildAttributesFromMetadata = async (
       extendedRequestId,
       requestId,
       retryCount: attempts,
+      data: res?.output,
       request: {
         operation,
+        params: commandInput,
         httpRequest: {
           region,
           statusCode,
@@ -94,11 +97,13 @@ function addFlags(http: HttpResponse, subsegment: Subsegment, err?: SdkError): v
 const getXRayMiddleware = (config: RegionResolvedConfig, manualSegment?: SegmentLike): BuildMiddleware<any, any> => (next: any, context: any) => async (args: any) => {
   const segment = contextUtils.isAutomaticMode() ? contextUtils.resolveSegment() : manualSegment;
   const {clientName, commandName} = context;
-  const operation: string = commandName.slice(0, -7); // Strip trailing "Command" string
+  const { input: commandInput } = args;
+  const commandOperation: string = commandName.slice(0, -7); // Strip trailing "Command" string
+  const operation: string = commandOperation.charAt(0).toLowerCase() + commandOperation.slice(1);
   const service: string = clientName.slice(0, -6);    // Strip trailing "Client" string
 
   if (!segment) {
-    const output = service + '.' + operation.charAt(0).toLowerCase() + operation.slice(1);
+    const output = service + '.' + operation;
 
     if (!contextUtils.isAutomaticMode()) {
       logger.getLogger().info('Call ' + output + ' requires a segment object' +
@@ -134,6 +139,7 @@ const getXRayMiddleware = (config: RegionResolvedConfig, manualSegment?: Segment
       service,
       operation,
       await config.region(),
+      commandInput,
       res,
       null,
     );
@@ -150,6 +156,7 @@ const getXRayMiddleware = (config: RegionResolvedConfig, manualSegment?: Segment
         service,
         operation,
         await config.region(),
+        commandInput,
         null,
         err,
       );


packages/core/lib/patchers/aws3_p.ts
```diff
@@ -40,6 +40,7 @@ const buildAttributesFromMetadata = async (
   service: string,
   operation: string,
   region: string,
+  commandInput: any,
```
carolabadeer, line 43: Is it possible to modify this to account for the possibility of the `Command` not having an `input` property?

carolabadeer (changes requested):
Look good so far! I think adding some [unit tests](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/test/unit/patchers/aws3_p.test.js) to cover this new attribute would be helpful as well

packages/core/lib/patchers/aws3_p.ts
SergeyPoluektov: AFAIK all the commands from AWS SDK v3 inherits from the abstract class `Command` from internal package [`smithy-client`](https://github.com/aws/aws-sdk-js-v3/blob/main/packages/smithy-client/src/command.ts). Meaning there shouldn't be a case when the `Command` doesn't have an `input` prop.

But, ofc, we can be a bit more defensive – introduce a few conditions and fallback to a previous behavior.

SergeyPoluektov:
> Look good so far! I think adding some [unit tests](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/test/unit/patchers/aws3_p.test.js) to cover this new attribute would be helpful as well

Thank you for review @carolabadeer I'll add tests for the attribute, sure.

scorsi:
Unfortunately, I tried your fix @SergeyPoluektov and it seems that it don't works well. I'm requesting on the exact same table and both node are created from that. There are get (working for sure), put (not working for sure) and delete (don't really know) operations into that Lambda. 😭
Capture d’écran 2022-09-12 à 20 33 54
The patch file I created :
```patch
diff --git a/node_modules/aws-xray-sdk-core/dist/lib/patchers/aws3_p.js b/node_modules/aws-xray-sdk-core/dist/lib/patchers/aws3_p.js
index 74a7aa5..935d15e 100644
--- a/node_modules/aws-xray-sdk-core/dist/lib/patchers/aws3_p.js
+++ b/node_modules/aws-xray-sdk-core/dist/lib/patchers/aws3_p.js
@@ -13,15 +13,17 @@ const logger = require('../logger');
 const { safeParseInt } = require('../utils');
 const utils_1 = require("../utils");
 const XRAY_PLUGIN_NAME = 'XRaySDKInstrumentation';
-const buildAttributesFromMetadata = async (service, operation, region, res, error) => {
+const buildAttributesFromMetadata = async (service, operation, region, commandInput, res, error) => {
     var _a, _b, _c;
     const { extendedRequestId, requestId, httpStatusCode: statusCode, attempts } = ((_a = res === null || res === void 0 ? void 0 : res.output) === null || _a === void 0 ? void 0 : _a.$metadata) || (error === null || error === void 0 ? void 0 : error.$metadata);
     const aws = new aws_1.default({
         extendedRequestId,
         requestId,
         retryCount: attempts,
+        data: res?.output,
         request: {
             operation,
+            params: commandInput,
             httpRequest: {
                 region,
                 statusCode,
@@ -60,10 +62,12 @@ function addFlags(http, subsegment, err) {
 const getXRayMiddleware = (config, manualSegment) => (next, context) => async (args) => {
     const segment = contextUtils.isAutomaticMode() ? contextUtils.resolveSegment() : manualSegment;
     const { clientName, commandName } = context;
-    const operation = commandName.slice(0, -7); // Strip trailing "Command" string
+    const { input: commandInput } = args;
+    const commandOperation = commandName.slice(0, -7); // Strip trailing "Command" string
+    const operation = commandOperation.charAt(0).toLowerCase() + commandOperation.slice(1);
     const service = clientName.slice(0, -6); // Strip trailing "Client" string
     if (!segment) {
-        const output = service + '.' + operation.charAt(0).toLowerCase() + operation.slice(1);
+        const output = service + '.' + operation;
         if (!contextUtils.isAutomaticMode()) {
             logger.getLogger().info('Call ' + output + ' requires a segment object' +
                 ' passed to captureAWSv3Client for tracing in manual mode. Ignoring.');
@@ -88,7 +92,7 @@ const getXRayMiddleware = (config, manualSegment) => (next, context) => async (a
         if (!res) {
             throw new Error('Failed to get response from instrumented AWS Client.');
         }
-        const [aws, http] = await buildAttributesFromMetadata(service, operation, await config.region(), res, null);
+        const [aws, http] = await buildAttributesFromMetadata(service, operation, await config.region(), commandInput, res, null);
         subsegment.addAttribute('aws', aws);
         subsegment.addAttribute('http', http);
         addFlags(http, subsegment);
@@ -97,7 +101,7 @@ const getXRayMiddleware = (config, manualSegment) => (next, context) => async (a
     }
     catch (err) {
         if (err.$metadata) {
-            const [aws, http] = await buildAttributesFromMetadata(service, operation, await config.region(), null, err);
+            const [aws, http] = await buildAttributesFromMetadata(service, operation, await config.region(), commandInput, null, err);
             subsegment.addAttribute('aws', aws);
             subsegment.addAttribute('http', http);
             addFlags(http, subsegment, err);
```

willarmiros:
Hi @SergeyPoluektov - any chance you could follow up with the unit tests?

andres-sinapsis:
Any news or progress about this PR? We really need to visualize the nodes by resource.

burakince:
Hi @SergeyPoluektov and @carolabadeer ,

Any update on this PR? Do you need any help? We really like to have this feature.

Best,

carolabadeer:
@SergeyPoluektov can you follow up with unit tests? If not, we are open to any contributions to this PR.

bilalq:
This really feels like something AWS themselves should put resources behind instead of deferring to community volunteers. X-Ray is a paid service (an expensive one at that) and now that AWS SDK v2 is going into maintenance mode, this is essentially a service degradation.

SchollSimon:
I totally agree this better be up and running fast. That is such a nogo first AWS encouraging developers to move to AWS SDKv3 then not delivering on the core features. We need this information for tracing distributed software systems it basically doesn't add any value at all at the moment.

danno-s:
Hi, commenting here for visibility:
Since this PR looks abandoned and this change is important to maintain feature parity with the AWS SDK v2, I took the changes made here and added unit tests in #611 to validate that command inputs and outputs get mapped correctly into segments.

carolabadeer:
closing as #611, which builds on the implementation in this PR, has been merged

PR #529, opened 2022-08-17
Fix prototype pollution issue

carolabadeer:
*Issue #, if available:*

*Description of changes:*
This change fixes the possible prototype pollution in the `addMetadata` method in segment.js and subsegment.js and adds corresponding unit tests.

### DIFF
--- a/packages/core/lib/segments/attributes/subsegment.js
+++ b/packages/core/lib/segments/attributes/subsegment.js
@@ -172,7 +172,9 @@ Subsegment.prototype.addMetadata = function(key, value, namespace) {
     this.metadata[ns] = {};
   }
 
-  this.metadata[ns][key] = value !== null && value !== undefined ? value : '';
+  if (ns !== '__proto__') {
+    this.metadata[ns][key] = value !== null && value !== undefined ? value : '';
+  }
 };
 
 Subsegment.prototype.addSqlData = function addSqlData(sqlData) {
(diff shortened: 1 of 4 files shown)


packages/core/lib/segments/attributes/subsegment.js
```diff
@@ -172,7 +172,9 @@ Subsegment.prototype.addMetadata = function(key, value, namespace) {
     this.metadata[ns] = {};
   }
 
-  this.metadata[ns][key] = value !== null && value !== undefined ? value : '';
+  if (ns !== '__proto__') {
```
willarmiros, line 175: Do we want to check for the string literal `'__proto__'` or the object `__proto__`?

E.g. what would happen if I tried to call `segment.addMetadata('key', 'value', __proto__)`?

carolabadeer: That's a great callout and a case I tested as well! 

On line 159 above there's a check that enforces a string type on the namespace object:
 `if (namespace && typeof namespace !== 'string')` 
 in which case an error would be logged and the method returns. This ensures that if `__proto__ `gets passed in, which is not a string, this line would never be reached and thus would not be an issue.

willarmiros: Got it! Thanks for explaining

willarmiros (approved):
lgtm

PR #541, opened 2022-10-14
Oversampling Mitigation 

carolabadeer:
*Issue #, if available:* N/A

*Description of changes:* Implemented oversampling mitigation - added methods to create subsegments without sampling (control sampling decisions at the subsegment level) and an SQS message helper class 

### DIFF
--- a/packages/core/lib/env/sqs_message_helper.js
+++ b/packages/core/lib/env/sqs_message_helper.js
@@ -1,0 +1,17 @@
+var LambdaUtils = require('../utils').LambdaUtils;
+
+class SqsMessageHelper {
+
+    isSampled(message){ 
+        const {attributes} = message; // extract attributes from message 
+
+        if (!'AWSTraceHeader' in attributes){
+            return false; 
+        }
+
+        return attributes['AWSTraceHeader'].includes('Sampled=1');
+    }
+
+}
+
+export default SqsMessageHelper;
--- a/packages/core/lib/patchers/aws_p.js
+++ b/packages/core/lib/patchers/aws_p.js
@@ -81,7 +81,8 @@ function captureAWSRequest(req) {
 
   var buildListener = function(req) {
     req.httpRequest.headers['X-Amzn-Trace-Id'] = 'Root=' + traceId + ';Parent=' + subsegment.id +
-      ';Sampled=' + (subsegment.segment.notTraced ? '0' : '1');
+      ';Sampled=' + (subsegment.isSampled ? '1' : '0');
+      
   };
 
   var completeListener = function(res) {
--- a/packages/core/lib/segments/attributes/subsegment.js
+++ b/packages/core/lib/segments/attributes/subsegment.js
@@ -28,6 +28,7 @@ Subsegment.prototype.init = function init(name) {
   this.start_time = SegmentUtils.getCurrentTime();
   this.in_progress = true;
   this.counter = 0;
+  this.isSampled = true;
 };
 
 /**
@@ -38,8 +39,22 @@ Subsegment.prototype.init = function init(name) {
 
 Subsegment.prototype.addNewSubsegment = function addNewSubsegment(name) {
   var subsegment = new Subsegment(name);
+  subsegment.isSampled = true;
   this.addSubsegment(subsegment);
   return subsegment;
+};
+
+Subsegment.prototype.addSubsegmentWithoutSampling = function addSubsegmentWithoutSampling(subsegment){
+  subsegment.isSampled = false;
+  this.addSubsegment(subsegment);
+};
+
+Subsegment.prototype.addNewSubsegmentWithoutSampling = function addNewSubsegmentWithoutSampling(name){
+  var subsegment = new Subsegment(name);
+  subsegment.isSampled = false;
+  this.addSubsegment(subsegment);
+  
+  return subsegment; 
 };
 
 /**
@@ -60,11 +75,14 @@ Subsegment.prototype.addSubsegment = function(subsegment) {
   subsegment.segment = this.segment;
   subsegment.parent = this;
 
-  if (subsegment.end_time === undefined) {
+  if (subsegment.end_time === undefined && subsegment.isSampled) {
     this.incrementCounter(subsegment.counter);
   }
-
-  this.subsegments.push(subsegment);
+  // don't push to subsegment array if subsegment is not sampled
+  if(subsegment.isSampled){ 
+    this.subsegments.push(subsegment);
+  }
+  
 };
 
 /**
@@ -286,7 +304,7 @@ Subsegment.prototype.close = function close(err, remote) {
   }
 
   if (root && root.counter > SegmentUtils.getStreamingThreshold()) {
-    if (this.streamSubsegments() && this.parent) {
+    if (this.streamSubsegments() && this.parent) { // this.isSampled ??
       this.parent.removeSubsegment(this);
     }
   }
@@ -362,7 +380,8 @@ Subsegment.prototype.streamSubsegments = function streamSubsegments() {
     var open = [];
 
     this.subsegments.forEach(function(child) {
-      if (!child.streamSubsegments()) {
+      if (!child.streamSubsegments()) { 
+        //&& child.isSampled?
         open.push(child);
       }
     });
--- a/packages/core/test/unit/segments/segment.test.js
+++ b/packages/core/test/unit/segments/segment.test.js
@@ -262,6 +262,57 @@ describe('Segment', function() {
     });
   });
 
+  describe('#addSubsegmentWithoutSampling', function (){
+    it('should have isSampled flag set to false', function(){
+      var segment = new Segment('parent');
+      var child = new Subsegment('child');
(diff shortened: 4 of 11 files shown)


packages/core/test/unit/segments/segment.test.js
```diff
@@ -262,6 +262,57 @@ describe('Segment', function() {
...
+      var child2 = new Subsegment('child-2');
+      segment.addSubsegmentWithoutSampling(child);
+      child.addSubsegmentWithoutSampling(child2)
+
+      assert.equal(child.isSampled, false);
+      assert.equal(child2.isSampled, false);
+    })
+
+    it('should not sample subsegment or subsegment of subsegment - mix', function(){
+      var segment = new Segment('parent');
```
atshaw43, line 302: Us needing to create the segment here signals that we re not running through the lambda specific code.

When the code thinks it is in lambda it will create a facade segment for us. The facade segment's trace data is filled out by what is in this variable. You can see an example in aws_lambda.test.js

process.env._X_AMZN_TRACE_ID
packages/core/test/unit/env/aws_lambda.test.js

I believe you need to call "init" here with that variable set.
packages/core/lib/env/aws_lambda.js

atshaw43:
Was this tested in Lambda directly?

packages/core/test/unit/segments/segment.test.js
```diff
@@ -262,6 +262,57 @@ describe('Segment', function() {
...
+      segment.addSubsegmentWithoutSampling(child);
+      child.addSubsegment(child2)
+      child.addSubsegmentWithoutSampling(child3)
+
+      assert.equal(child.isSampled, false);
+      assert.equal(child2.isSampled, true);
+      assert.equal(child3.isSampled, false);
+    })
+  });
+
```
atshaw43, line 315: After we close the segment, is it possible to check if the emitter sent out the segment or not?
And can we examine the trace header of the subsegment that is sent downstream to see what the sample flag is set to?

atshaw43 (changes requested):

packages/core/test/unit/segments/segment.test.js
```diff
@@ -262,6 +262,57 @@ describe('Segment', function() {
     });
   });
 
+  describe('#addSubsegmentWithoutSampling', function (){
```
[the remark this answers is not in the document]
carolabadeer: Thank you!

packages/core/test/unit/segments/segment.test.js
carolabadeer: Thanks for pointing that out! I've incorporated that change into the updated unit tests

carolabadeer: To answer the question above, yes, this was tested in Lambda using the sample sample JSON data used in the Java oversampling mitigation PR. 

For trace header with sampled=1:
image (3)

For trace header with sampled=0:
image (4)

packages/core/lib/segments/attributes/subsegment.js
```diff
@@ -38,10 +39,24 @@ Subsegment.prototype.init = function init(name) {
 
 Subsegment.prototype.addNewSubsegment = function addNewSubsegment(name) {
   var subsegment = new Subsegment(name);
+  subsegment.isSampled = true;
```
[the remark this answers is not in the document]
carolabadeer: Great callout! I believe checking the value of `subsegment.segment.notTraced` would be more accurate in this case as to capture the parent _segment's_ sampling decision not the parent _subsegment_ (since it is possible to have subsegments of subsegments.

packages/core/lib/patchers/aws_p.js
```diff
@@ -81,7 +81,8 @@ function captureAWSRequest(req) {
 
   var buildListener = function(req) {
     req.httpRequest.headers['X-Amzn-Trace-Id'] = 'Root=' + traceId + ';Parent=' + subsegment.id +
-      ';Sampled=' + (subsegment.segment.notTraced ? '0' : '1');
+      ';Sampled=' + (subsegment.isSampled ? '1' : '0');
```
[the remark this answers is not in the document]
willarmiros: +1

packages/core/lib/segments/attributes/subsegment.js
```diff
@@ -39,9 +40,23 @@ Subsegment.prototype.init = function init(name) {
...
   return subsegment;
 };
 
+Subsegment.prototype.addSubsegmentWithoutSampling = function addSubsegmentWithoutSampling(subsegment){
+  subsegment.isSampled = false;
+  this.addSubsegment(subsegment);
+};
+
+Subsegment.prototype.addNewSubsegmentWithoutSampling = function addNewSubsegmentWithoutSampling(name){
+  var subsegment = new Subsegment(name);
```
willarmiros, line 53: Nit: use `const`

packages/core/lib/env/sqs_message_helper.js
```diff
@@ -0,0 +1,14 @@
+class SqsMessageHelper {
+  constructor() {
+  }
+
+  isSampled(message) {
```
willarmiros, line 5: Can we make this `static`? Would need to update docs as well

packages/core/lib/context_utils.js
```diff
@@ -42,11 +42,11 @@ var contextUtils = {
    * @alias module:context_utils.resolveManualSegmentParams
    */
 
-  resolveManualSegmentParams: function resolveManualSegmentParams(params) {
+  resolveManualSegmentParams: function resolveManualSegmentParams (params) {
```
atshaw43, line 45: Confirmed with Carol,
The linter has not been run in a long time which is causing the large code churn.

packages/core/lib/patchers/aws3_p.ts
```diff
@@ -110,15 +110,21 @@ const getXRayMiddleware = (config: RegionResolvedConfig, manualSegment?: Segment
     return next(args);
   }
 
-  const subsegment: Subsegment = segment.addNewSubsegment(service);
+  let subsegment: Subsegment;
+  if (segment.notTraced == false || !segment.subsegments[segment.subsegments.length - 1].notTraced) {
```
atshaw43, line 114: Would this also work?
if (!segment.subsegments[segment.subsegments.length - 1].notTraced)

packages/core/lib/patchers/aws_p.js
```diff
@@ -73,22 +73,29 @@ function captureAWSRequest(req) {
...
+  const throttledError = this.throttledError || throttledErrorDefault;
 
-  var stack = (new Error()).stack;
-  var subsegment = parent.addNewSubsegment(this.serviceIdentifier);
-  var traceId = parent.segment ? parent.segment.trace_id : parent.trace_id;
+  const stack = (new Error()).stack;
 
-  var buildListener = function(req) {
+  let subsegment;
+  if (parent.notTraced === false || !parent.subsegments[parent.subsegments.length - 1].notTraced) {
```
atshaw43, line 81: Same here. Can we just look at the subsegment and not have the segment check?

packages/core/test/unit/patchers/aws_p.test.js
```diff
@@ -252,9 +252,93 @@ describe('AWS patcher', function() {
...
+    });
+
+    it('should log an info statement and exit if parent is not found on the context or on the call params', function (done) {
+      stubResolve.returns();
+      const logStub = sandbox.stub(logger, 'info');
+
+      awsClient.call(awsRequest);
+
+      setTimeout(function () {
+        logStub.should.have.been.calledOnce;
```
atshaw43, line 327: If it is not supposed to sample, then should this have emitted?

carolabadeer: I believe this is unrelated - just checking if a log is produced. There is another test right below checking the trace header being sent out

packages/core/lib/patchers/aws3_p.ts
carolabadeer: Yes it would! Will incorporate that change

carolabadeer: The reason this is being kept in is to account for subsegments that only have a parent segment (no parent subsegments). After some debugging, I discovered that this line 

`const segment = contextUtils.isAutomaticMode() ? contextUtils.resolveSegment() : manualSegment;`

always returns the **currently active segment**. If there are no active subsegments and the parent segment is unsampled, this check will not behave correctly without the `segment.notTraced == false`

packages/core/lib/patchers/aws_p.js
carolabadeer: See above

packages/core/lib/patchers/aws3_p.ts
carolabadeer: This is more of a design decision: check the parent sampling decision here for unsampled segments, or only check the subsegment and rely on the emitter not sending the entire segment if it is unsampled down the line. If the latter was used in the other SDKs, I am happy to incorporate the same design here as well.

willarmiros (approved):
LGTM - please wait for @atshaw43 review too

atshaw43 (approved):

PR #549, opened 2022-12-02
Propagate additional trace data into AWS requests on Lambda

srprash:
This PR allows the SDK to parse any additional key-value pair supplied in the X-Ray trace header. And then these additional key-value pairs are injected into instrumented AWS requests made from a Lambda function.

The purpose of this PR is similar to what [I did for the Python SDK](https://github.com/aws/aws-xray-sdk-python/pull/362). The difference there was that the segment already had functionality to save the original trace header. Here, I added a new property on the segment `additionalTraceData` which stores only the additional key-value pairs instead of the complete trace header.


.gitignore
```diff
@@ -5,3 +5,4 @@ docs
 .nyc_output/
 *.lcov
 dist
+*package-lock.json
```
carolabadeer, line 8: I believe we need the `package-lock.json` file to keep track of the versions of the packages we use, and it is also one of the files that is modified when we are preparing for a new release. Is there a reason for adding it to `.gitignore`?

packages/core/lib/patchers/aws_p.js
```diff
@@ -85,10 +85,17 @@ function captureAWSRequest(req) {
   }
 
   var traceId = parent.segment ? parent.segment.trace_id : parent.trace_id;
+  const data = parent.segment ? parent.segment.additionalTraceData : parent.additionalTraceData;
 
   var buildListener = function(req) {
-    req.httpRequest.headers['X-Amzn-Trace-Id'] = 'Root=' + traceId + ';Parent=' + subsegment.id +
+    let traceHeader = 'Root=' + traceId + ';Parent=' + subsegment.id +
```
carolabadeer, line 91: nit: maybe rename this variable to match the name below (`additionalTraceData`) for clarity on what this data is?

packages/core/lib/patchers/aws_p.js
```diff
@@ -85,10 +85,17 @@ function captureAWSRequest(req) {
...
       ';Sampled=' + (subsegment.notTraced ? '0' : '1');
+    if (data != null) {
```
carolabadeer, line 93: `!==` is a better alternative to `!=` in this case since it performs a more [strict type check ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_inequality)

packages/core/lib/utils.js
```diff
@@ -162,33 +162,37 @@ var utils = {
      */
     populateTraceData: function(segment, xAmznTraceId) {
       logger.getLogger().debug('Lambda trace data found: ' + xAmznTraceId);
-      var data = utils.processTraceData(xAmznTraceId);
+      var traceData = utils.processTraceData(xAmznTraceId);
```
carolabadeer, line 165: nit: use `let` instead of `var`

packages/core/lib/utils.js
```diff
@@ -217,19 +222,22 @@ var utils = {
       var pair = header.split('=');
 
       if (pair[0] && pair[1]) {
-        var key = pair[0].trim().toLowerCase();
-        var value = pair[1].trim().toLowerCase();
-        var reserved = reservedKeywords.indexOf(key) !== -1;
+        var key = pair[0].trim();
+        var value = pair[1].trim();
+        var lowerCaseKey = key.toLowerCase();
+        var reserved = reservedKeywords.indexOf(lowerCaseKey) !== -1;
```
carolabadeer, line 228: Is it necessary to create a new `lowerCaseKey` variable instead of converting the key & value to lowercase using `.toLowerCase`?

nit: use `let` instead of `var`.

srprash: I don't want to mutate the casing of additional trace data since it may contain a case-sensitive key. So I created a separate variable `lowerCaseKey` from `key` as the former is used in multiple places below for existing logic.
Also, I don't think we need to change the casing of the value anywhere.

packages/core/lib/patchers/aws_p.js
srprash: This is not just the additional trace data, rather the complete trace header string, that's why the name `traceHeader`.

packages/core/lib/patchers/aws_p.js
srprash: Does the type checking matter in JS when comparing with `null`? My understanding is that the expression `data != null` would return `false` when the `data` is either `null` or `undefined`. Whereas, `data !== null` only returns `false` when data is specifically `null`. Wouldn't that make `!=` more appropriate when comparing with `null`?

carolabadeer: Got it, thanks for the clarification! If the possible values we are checking for are null/undefined, then` !=` is the way to go.

carolabadeer (approved):
LGTM

packages/core/lib/patchers/aws_p.js
srprash: Yup. We intend to handle `undefined` and `null` the same way here.

Issue #563, opened 2023-01-26
Feature Request:  AWS_XRAY_COLLECT_SQL_QUERIES support for MySQL


jasonterando:
Support collection of SQL queries when AWS_XRAY_COLLECT_SQL_QUERIES is set for MySQL, similar to [PR 390](https://github.com/aws/aws-xray-sdk-node/pull/390).

PR #564, opened 2023-01-26
Add support for including sql query in sql subsegment for MySQL

jasonterando:
Issue: [563](https://github.com/aws/aws-xray-sdk-node/issues/563)

Description of changes: Add support for adding sql query to sql subsegment when AWS_XRAY_COLLECT_SQL_QUERIES is set.

### DIFF
--- a/packages/mysql/lib/mysql_p.js
+++ b/packages/mysql/lib/mysql_p.js
@@ -242,9 +242,7 @@ function captureOperation(name) {
       if (isPromise(command)) {
         command.then(() => {
           subsegment.close();
-        });
-
-        command.catch(errorCapturer);
+        }).catch (errorCapturer);
       } else {
         command.on('end', function() {
           subsegment.close();
@@ -254,16 +252,15 @@ function captureOperation(name) {
       }
     }
 
-    subsegment.addSqlData(createSqlData(config, command, args.sql));
+    subsegment.addSqlData(createSqlData(config, args.values, args.sql));
     subsegment.namespace = 'remote';
 
     return command;
   };
 }
 
-function createSqlData(config, command, sql) {
-  var commandType = command.values ? PREPARED : null;
-
+function createSqlData(config, values, sql) {
+  var commandType = values ? PREPARED : null;
   var data = new SqlData(DATABASE_VERS, DRIVER_VERS, config.user,
     config.host + ':' + config.port + '/' + config.database,
     commandType);
--- a/packages/mysql/lib/mysql_p.js
+++ b/packages/mysql/lib/mysql_p.js
@@ -254,19 +254,23 @@ function captureOperation(name) {
       }
     }
 
-    subsegment.addSqlData(createSqlData(config, command));
+    subsegment.addSqlData(createSqlData(config, command, args.sql));
     subsegment.namespace = 'remote';
 
     return command;
   };
 }
 
-function createSqlData(config, command) {
+function createSqlData(config, command, sql) {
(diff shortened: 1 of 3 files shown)


jasonterando:
Hi, updated to include the fixes to set SqlData values from arguments as opposed to the results from query/execute (which may be a promise and not have values) and to fix a bug when promises are returned by query/execute and the promise rejects.

packages/mysql/lib/mysql_p.js
```diff
@@ -254,19 +252,22 @@ function captureOperation(name) {
...
     subsegment.namespace = 'remote';
 
     return command;
   };
 }
 
-function createSqlData(config, command) {
-  var commandType = command.values ? PREPARED : null;
-
+function createSqlData(config, values, sql) {
```
carolabadeer, line 262: Hi, thanks for this contribution! 
You mentioned this issue + the fix for it in the bug report, maybe keeping `createSqlData(config, args)` with 2 parameters might be better to remain consistent with the [same function in postgres](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/postgres/lib/postgres_p.js#L125)?

packages/mysql/test/unit/mysql_p.test.js
```diff
@@ -141,7 +141,7 @@ describe('captureMySQL', function() {
         var stubDataInit = sandbox.stub(SqlData.prototype, 'init');
         var config = conn.config;
 
-        query.call(connectionObj, 'sql here');
+        query.call(connectionObj, 'sql here', [1]);
```
carolabadeer, line 144: Just trying to get a better understanding of this, do you mind clarifying the purpose of passing in `[1]` here?

packages/mysql/lib/mysql_p.js
jasonterando: Well... other than the number of arguments, the calls to createSqlData are not very consistent in the current state, primarily because the value getting passed in to createSqlData in mysql_p can either be a command or a _promise_ to return results (which is the whole reason this PR exists in the first place).

As far as the number of arguments to createSqlData in mysql_p.js, I could pass in the args which consistently contains both the SQL and parameters, which would keep it at two arguments; albeit the second argument by necessity is different than what gets passed in to createSqlData in postgres_p.js.  In untyped languages I always found it easier to explicitly accept discrete values as opposed to a composited object, but it's a nit.  If two arguments is important, I can make that work and resubmit so I can log SQL with MySQL in XRay.

carolabadeer: Yea they definitely do different things, this was just a nit pick to target as much consistency as possible. There isn't a strong preference for either option on my end, so I'll leave that decision up to you. I think a small comment explaining how/why this function is different can also be very helpful for anyone looking back on this implementation

carolabadeer (approved):

packages/mysql/lib/mysql_p.js
jasonterando: Ok, added function notes per your request

jasonterando:
Ok, I give up... I don't know what Lerna is, and I do not have the time to learn.  Literally the same code was working two weeks ago with only an additional comment added.

carolabadeer:
> Ok, I give up... I don't know what Lerna is, and I do not have the time to learn. Literally the same code was working two weeks ago with only an additional comment added.

@jasonterando apologies for the confusion. This is a CI issue that is not related to your changes, I'll be looking into it and will merge this PR once it's fixed

PR #572, opened 2023-02-13
remove html5shiv to support legacy IE versions

srprash:
Pulling in this script for HTML5 shiv to support legacy IE version has been raising security concerns internally. I think we can remove it from creeping into our generated docs.


willarmiros (approved):
Nice!

PR #591, opened 2023-06-02
Add http field in Segment and Subsegment type

hermanho:
*Issue #, if available:*

*Description of changes:*
Add http field in Segment and Subsegment type


packages/core/lib/segments/attributes/subsegment.d.ts
```diff
@@ -12,6 +13,8 @@ declare class Subsegment {
   namespace?: string;
   notTraced: boolean;
 
+  http?: IncomingRequestData;
```
carolabadeer, line 16: I believe the `http` type definition here should have type `RemoteRequestData`, not `IncomingRequestData`, based on the [logic in subsegment.js where this.http is being assigned to a new RemoteRequestData object](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/segments/attributes/subsegment.js#L253). This is different from the behavior in [segment.js where this.http is assigned to a new IncomingRequestData object](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/segments/segment.js#L67-L74).

Issue #61, opened 2018-10-19
Adding an error to a sub/segment, forces a fault flag.


Tanbouz:
**Details**

`this.addFaultFlag();` 

https://github.com/aws/aws-xray-sdk-node/blob/b8996144eb28293c436667e9cee9320938e8e5da/packages/core/lib/segments/segment.js#L228-L234

[AWS X-ray concepts regarding error categorization](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-errors)

> X-Ray tracks errors that occur in your application code, and errors that are returned by downstream services. Errors are categorized as follows.
> 
> Error – Client errors (400 series errors)
> Fault – Server faults (500 series errors)
> Throttle – Throttling errors (429 Too Many Requests)

**Issue**

if `addError` refers directly to client errors then it is a bit confusing that it flags it as a "Fault" and not an "Error" flag. 

If `addError` is more of an abstract method for all **x-ray error types**, then I don't have the option to specifiy which kind of **x-ray error** it is per the categorizations above. And I'm forced to have it flagged as a "Fault".

Tanbouz:
It is no big deal. I'm new to X-Ray and wished to understand and use the terminology better.  

I simply wanted to use custom segments and sub-segments in manual mode to wrap sections of my functions and didn't like 😛 that I was forced to flag all "exceptions" as **Faults** because I wanted to categorize them like so:

1- Flag the exception as a fault if it something that is handled but I still wish to track. Something like a 3rd party service that fails (HTTP request). I can also see the stack trace on why it failed.
```
addError()
addFaultFlag()
```

2- Flag the exception as an unhandled error otherwise. Something needs to be handled or fixed due my own code logic/syntax.
```
addError()
addErrorFlag()
```

3- Flag it as throttling if I have my own throttling logic
```
addError()
addThrottleFlag()
```

4- Flag it as throttling and a fault if a 3rd party service throttles.
```
addError()
addThrottleFlag()
addFaultFlag()
```

5- Flag it as a fault and error if a 3rd party service fault is unhandled.
```
addError()
addErrorFlag()
addFaultFlag()
```

I can always implement my own tagging in the segment's metadata. So it is no issue I guess. I will leave it up to you 😄

haotianw465:
[...]

PR #593, opened 2023-06-10
Add support for specifying fault flag when adding error

icholy:
Issue https://github.com/aws/aws-xray-sdk-node/issues/61

*Description of changes:*

This change enables setting an error without setting the `fault` flag.

*Example:*

``` js
segment.addError(err, false, false);
```

### DIFF
--- a/packages/core/lib/segments/segment.js
+++ b/packages/core/lib/segments/segment.js
@@ -280,16 +280,19 @@ Segment.prototype.removeSubsegment = function removeSubsegment(subsegment) {
  * Adds error data into the segment.
  * @param {Error|string} err - The error to capture.
  * @param {boolean} [remote] - Flag for whether the exception caught was remote or not.
+ * @param {boolean} [fault] - Flag for whether the fault flag should be set.
  */
 
-Segment.prototype.addError = function addError(err, remote) {
+Segment.prototype.addError = function addError(err, remote, fault) {
   if (err == null || typeof err !== 'object' && typeof(err) !== 'string') {
     logger.getLogger().error('Failed to add error:' + err + ' to subsegment "' + this.name +
     '".  Not an object or string literal.');
     return;
   }
 
-  this.addFaultFlag();
+  if (fault || fault === undefined) {
+    this.addFaultFlag();
+  }
 
   if (this.exception) {
     if (err === this.exception.ex) {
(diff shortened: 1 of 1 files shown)


packages/core/lib/segments/segment.js
```diff
@@ -280,16 +280,19 @@ Segment.prototype.removeSubsegment = function removeSubsegment(subsegment) {
  * Adds error data into the segment.
  * @param {Error|string} err - The error to capture.
  * @param {boolean} [remote] - Flag for whether the exception caught was remote or not.
+ * @param {boolean} [fault] - Flag for whether the fault flag should be set.
  */
 
-Segment.prototype.addError = function addError(err, remote) {
+Segment.prototype.addError = function addError(err, remote, fault) {
```
carolabadeer, line 286: Changing this function signature will be a breaking / backwards-incompatible change. Ideally, we would want to maintain the current expected behaviour so that this change does not cause any unexpected changes for others as well. 

I would recommend setting a default value of `true` for the `fault` parameter to ensure that the fault flag is always added by default, and would only be excluded if a third argument with the value of `false` is included. 

I am also curious, what is the use case for excluding the fault flag? Is there a specific reason why it would not be included?

carolabadeer (changes requested):

packages/core/lib/segments/segment.js
rafael-castelo: @carolabadeer  Here I would like to answer, as I am having the same need as @icholy. In the segments error [documentation](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html#api-segmentdocuments-errors), is mentioned that the flags `error`, `throttle` and `fault`, can represent different types of errors. 
In my personal use case, we are flagging all expected errors as `error` and unexpected errors as `fault`.
At the moment of catching and handling the error logic, we want to add error data to the segment with `addError` method, but it always set `fault` flag to `true`. So we need to overwrite the flag accordingly, whether the error is expected or unexpected.

An extra param as @icholy suggest will reduce extra code, and avoid overwriting the segment properties directly.

icholy: @carolabadeer the change _is_ backwards compatible. The `fault || fault === undefined` condition is semantically equivalent to defaulting `fault` to true.

PR #595, opened 2023-06-12
fix #594 memory leaks caused by cls-hooked

### LINKED ISSUE #594: memory leaks caused by cls-hooked
Hi,

We noticed a massive memory leak when using this library. We managed to find that the memory leaks come from the unmaintained cls-hooked library.
Since we are using the library in manual mode, we don't even want it to be used, but currently, you create a namespace by default. When setting manual mode, the namespace gets deleted, but cls-hooks doesn't do that properly, creating a memory leak.

A proposed solution will be to not use cls if an env variable exists to control manual mode.

PR with the fix is submitted

regevbr:
*Issue #594

*Description of changes:*

Add option to set manual mode from env variable `AWS_XRAY_MANUAL_MODE`

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -26,6 +26,8 @@ By default, the SDK is in automatic mode. You can flip the mode of the SDK using
 
     AWSXRay.enableAutomaticMode();
 
+    process.env.AWS_XRAY_MANUAL_MODE
+
 #### Automatic mode
 
 Automatic mode is designed for use with Express, Restify, and Lambda
@@ -54,6 +56,7 @@ section for different usages.
 **Environment variables always override values set in code.**
 
     AWS_XRAY_DEBUG_MODE              Enables logging of debug messages to console output. Logging to a file is no longer built in. See 'configure logging' below.
+    AWS_XRAY_MANUAL_MODE             For overriding the default automatic mode. See 'Automatic mode'.
     AWS_XRAY_TRACING_NAME            For overriding the default segment name to use
     with the middleware. See 'dynamic and fixed naming modes'.
     AWS_XRAY_DAEMON_ADDRESS          For setting the daemon address and port.
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -22,11 +22,15 @@ mechanisms, but a few are supplied. See Capturing Function Calls below.
 The AWS X-Ray SDK has two modes: `manual` and `automatic`.
 By default, the SDK is in automatic mode. You can flip the mode of the SDK using the following:
 
-    AWSXRay.enableManualMode();
+```js
+AWSXRay.enableAutomaticMode();
 
-    AWSXRay.enableAutomaticMode();
+AWSXRay.enableManualMode();
 
-    process.env.AWS_XRAY_MANUAL_MODE
+process.env.AWS_XRAY_MANUAL_MODE = 'true';
+```
+
+see [ #595](https://github.com/aws/aws-xray-sdk-node/pull/595) for details
 
 #### Automatic mode
 
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -27,10 +27,12 @@ AWSXRay.enableAutomaticMode();
 
 AWSXRay.enableManualMode();
 
+/* see https://github.com/aws/aws-xray-sdk-node/pull/595
+ for details on using this environment variable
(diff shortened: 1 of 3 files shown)


packages/core/lib/context_utils.js
```diff
@@ -205,8 +205,13 @@ var contextUtils = {
...
 
-cls.createNamespace(NAMESPACE);
-logger.getLogger().debug('Starting the AWS X-Ray SDK in automatic mode (default).');
+if (process.env.AWS_XRAY_MANUAL_MODE) {
+  cls_mode = false;
+  logger.getLogger().debug('Starting the AWS X-Ray SDK in manual mode.');
+} else {
+  cls.createNamespace(NAMESPACE);
+  logger.getLogger().debug('Starting the AWS X-Ray SDK in automatic mode (default).');
+}
```
carolabadeer, line 214: Is it possible to refactor this change to not rely on a new env variable (`AWS_XRAY_MANUAL_MODE`)?Possibly using just the `cls_mode` variable?

regevbr: I wish, but this is done on the module load, which doesn't give a chance to the developer to change the `cls_mode` value.
To do it at a later stage will mean a breaking change. I'm willing to implement that, but I need to know that this is ok with you and how do you want to tackle that change

carolabadeer: Do you mind going into more detail about what the change would be in that case and why/how it would be a breaking change? Ideally, I would want to aim for a solution that fixes this issue out of the box without needing a change from the customer's end.

carolabadeer: Got it, thanks for the detailed information! In that case, I agree with your approach for the fix here :)

packages/core/README.md
```diff
@@ -26,6 +26,8 @@ By default, the SDK is in automatic mode. You can flip the mode of the SDK using
 
     AWSXRay.enableAutomaticMode();
 
+    process.env.AWS_XRAY_MANUAL_MODE
```
carolabadeer, line 29: Nit: Can this line be moved to be directly beneath `AWSXRay.enableManualMode()` and also include a comment with the link to this PR for context? This will be very helpful to anyone reading the documentation, especially if they are looking for a fix for this memory leak.

regevbr: I tried something else, can you suggest a better way to present it?

packages/core/README.md
```diff
@@ -22,9 +22,15 @@ mechanisms, but a few are supplied. See Capturing Function Calls below.
...
 By default, the SDK is in automatic mode. You can flip the mode of the SDK using the following:
 
-    AWSXRay.enableManualMode();
+```js
+AWSXRay.enableAutomaticMode();
+
+AWSXRay.enableManualMode();
+
+process.env.AWS_XRAY_MANUAL_MODE = 'true';
+```
```
carolabadeer, line 31: Maybe something like this to keep the comment directly above the `AWS_XRAY_MANUAL_MODE` env variable?
```suggestion
```js
AWSXRay.enableAutomaticMode();

AWSXRay.enableManualMode();

// see https://github.com/aws/aws-xray-sdk-node/pull/595 for details on using this environment variable to prevent memory leaks when using manual mode
process.env.AWS_XRAY_MANUAL_MODE = 'true';
```
```

File: packages/core/README.md [hunk 4]
```diff
@@ -22,9 +22,15 @@ mechanisms, but a few are supplied. See Capturing Function Calls below.
 The AWS X-Ray SDK has two modes: `manual` and `automatic`.
 By default, the SDK is in automatic mode. You can flip the mode of the SDK using the following:
 
-    AWSXRay.enableManualMode();
+```js
+AWSXRay.enableAutomaticMode();
+
+AWSXRay.enableManualMode();
+
+process.env.AWS_XRAY_MANUAL_MODE = 'true';
```
carolabadeer, line 30: Maybe something like this to keep the comment directly above the `AWS_XRAY_MANUAL_MODE` env variable?
```suggestion

// see https://github.com/aws/aws-xray-sdk-node/pull/595 for details on using this environment variable to prevent memory leaks when using manual mode
process.env.AWS_XRAY_MANUAL_MODE = 'true';
```

regevbr: done

carolabadeer (approved):
LGTM - thanks @regevbr!

PR #597, opened 2023-06-14
[CI] Keep lerna version consistent. Test on Node20. And more.

srprash:
### Problem
Recently lerna v7 got released and it broke our CI due to backward incompatible command. Upon investigation, I found that although we specify lerna v5.5.2 in our package.json, the latest version was being used for builds.

### Solution
Use the lerna version that is defined in the package.json. This is currently v5.5.2 which can be used for our build without any problems. In the future, we may think of migrating to a newer mechanism as per [this doc](https://lerna.js.org/docs/legacy-package-management).

### Changes
- Use `npm install` before using lerna to ensure that the one from our dependency is being used.
- Add testing on Node20
- Drop testing for Node14 on Windows due to [an issue with npm](https://github.com/npm/cli/issues/4341#issuecomment-1040608101) that the upstream will not fix.
- Minor refactors to the workflow, including moving lint to a separate workflow from build.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,40 @@
+name: Lint
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  lint:
+    name: Check code style
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


carolabadeer (approved):
LGTM!

Issue #599, opened 2023-06-23
Vulnerability in the library


rpodwika:
https://www.mend.io/vulnerability-database/CVE-2022-25883

-> aws-xray-sdk-3.5.0.tgz (Root Library)

   -> aws-xray-sdk-core-3.5.0.tgz

     -> cls-hooked-4.2.2.tgz

       -> :x: **semver-5.7.1.tgz** (Vulnerable Library)

jhonnycordova:
Hello. Are you planning to fix this? Any workaround I can use in the meantime?. Thanks

carolabadeer:
Hi @rpodwika and @jhonnycordova, thanks for raising this issue 

Do you mind clarifying if this vulnerability is originating from the `semver v7.3.8` in the [aws-xray-sdk-core package dependencies](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/package.json#L31) or `semver v6.3.0` being brought in from [cls-hooked](https://github.com/Jeff-Lewis/cls-hooked/blob/master/package.json#L44)? Where is `v5.7.1` being brought in?

kryten87:
The `cls-hooked` package on the `master` branch does indeed have `semver` v6.3.0, but the `v4.2.2` tag has `semver` v5.4.1. I'm not sure where 5.7.1 is coming from. In any event, the vulnerability described in the link above affects any version of `semver` prior to 7.5.2, so even installing from `master` will not resolve the problem.

The version from `cls-hooked` may be transitive, but it is enough to cause `npm audit` to complain.

cortexcompiler:
I am seeing this vulnerability flagged for any version of semver < 7.5.2:
https://security.snyk.io/vuln/SNYK-JS-SEMVER-3247795

The path given is:
> Introduced through: aws-xray-sdk-core@3.5.0 > cls-hooked@4.2.2 > semver@5.7.1

carolabadeer:
Hi all, thank you for your responses! We are actively working on a fix

carolabadeer:
Node SDK v3.5.1 which includes the fix for this security vulnerability has been released 
https://github.com/aws/aws-xray-sdk-node/releases/tag/aws-xray-sdk-node%403.5.1

PR #604, opened 2023-07-13
Override transitive semver dependency version to fix security vulnerability 

carolabadeer:
*Issue #, if available:* #599

*Description of changes:* Use `npm-force-resolutions` to override the vulnerable `semver` dependency `v5.7.1` being installed from `cls-hooked v4.2.2`. It is necessary to manually override the `semver` version as the `cls-hooked` package has not issued a release since 2017.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,9 @@
   "version": "3.5.0",
   "private": true,
   "license": "Apache-2.0",
+  "resolutions": {
+    "semver": "^7.5.3"
+  },
   "devDependencies": {
     "@aws-sdk/config-resolver": "^3.4.1",
     "@aws-sdk/middleware-stack": "^3.3.0",
@@ -55,6 +58,9 @@
   "engines": {
     "node": ">= 14.x",
     "npm": ">= 2.x"
+  },
+  "scripts": {
+    "preinstall": "npx npm-force-resolutions"
   },
   "dependencies": {
     "aws-xray-sdk": "file:packages/full_sdk",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "license": "Apache-2.0",
   "resolutions": {
-    "semver": "^7.5.3"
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -3,6 +3,9 @@
   "version": "3.5.0",
   "private": true,
   "license": "Apache-2.0",
+  "resolutions": {
+    "semver": "7.5.3"
```
wangzlei, line 7: wondering why is not 
```
"resolutions": {
    "cls-hooked" : {
        "semver": "7.5.3"
    }
}
```

wangzlei (approved):

srprash (approved):

srprash:
How do we ensure that `cls-hooked` works with the "forced" v7.5.3 of `semver` in our SDK and not behave unexpectedly?

carolabadeer:
@srprash `cls-hooked` only uses `semver` in [one line](https://github.com/Jeff-Lewis/cls-hooked/blob/v4.2.2/index.js#L11) to utilize the `gte` function, and the [module that exports the gte function in the semver repository ](https://github.com/npm/node-semver/blob/36cd334708ec1f85a71445622fb1864bceee0f4e/functions/gte.js#L1-L3) has not had any changes in the last 4 years. So the single usage of the `semver.gte()` function in `cls-hooked` would behave the same for the current version in `cls-hooked` (5.7.1) and the forced version we are using (7.5.3) since [5.7.1 was released 4 years ago](https://github.com/npm/node-semver/releases/tag/v5.7.0).

package.json
carolabadeer: I tried that initially but it seems that `npm-force-resolutions` doesn't support specifying the direct dependency -> transitive dependency. When I ran `lerna bootstrap --hoist` to install,  the `package-lock.json` was not updated correctly in this case.

wangzlei: Good to know

PR #607, opened 2023-07-18
Better dependency override. Npm 8 in workflows. 

srprash:
- Updated the workflows to use npm v8.
- As a better practice, switched to using [`npm ci`](https://docs.npmjs.com/cli/v9/commands/npm-ci) (clean install) instead of `npm install` in the workflows. This way we are not updating the `package-lock.json` on the fly, rather using the packages specified in the lock file.
- Updated to use `overrides` for explicitly bumping the `semver` version in `cls-hooked`. 
- Used npm v8 to generate newer `package-lock.json`. Verified that the `semver` version was bumped. 
- Now the SDK should be built using npm v8.


carolabadeer (approved):
Looks good to me!

PR #619, opened 2023-10-18
fix(core): serialize bigint in metadata to string

### LINKED ISSUE #616: Segment containing BigInt in metadata throws an error when serialized
Hi, one of our customers (Powertools for AWS (TypeScript)) reported (aws-powertools/powertools-lambda-typescript#1713) that the `addMetadata` method throws an error when the metadata object contains a `BigInt` value.

Example:
```ts
const segment = new Subsegment('## foo.bar');

segment.addMetadata('foo', {
  bar: BigInt(123),
});
```

BigInt is a native JavaScript object but cannot be serialized by the built-in `JSON.stringify()` function. Because of this, the SDK allows you to add the value as metadata but eventually throws an error when the segment is serialized right before being emitted.

The `JSON.stringify()` object allows a [`replacer`](https://tc39.es/ecma262/multipage/structured-data.html#sec-json.stringify) function as second argument, which could be used to encode this, as well as other exotic objects.

In this case, the change would be minimal and be applied to the `format()` and `toString()` methods of the Segment object, since they both use call `JSON.stringify()` [here](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/segments/segment.js#L434) and [here](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/segments/segment.js#L438).

After the change, the implementation would look something like this:

```diff
Subsegment.prototype.format = function format() {
    this.type = 'subsegment';
    if (this.parent) {
        this.parent_id = this.parent.id;
    }
    if (this.segment) {
        this.trace_id = this.segment.trace_id;
    }
--  return JSON.stringify(this);
++ return JSON.stringify(this, replacer());
}; 
/**
 * Returns the formatted subsegment JSON string.
 */
Subsegment.prototype.toString = function toString() {
--  return JSON.stringify(this);
++ return JSON.stringify(this, replacer());
};

++ function replacer(_, value) {
++   if (typeof value === 'bigint') {
++     return value.toString();
++   }
++
++   return value;
++ }
```

As mentioned, the change would be minimal and would impact only BigInt objects, for all other types the added code would be equivalent to a no-op.

As an alternative solution, if the replacer idea is not welcome, I would propose to extract the `JSON.stringify()` calls to its own method in the Segment object so that it can be easily overridden/extended by consumers, i.e. : 

```diff
Subsegment.prototype.format = function format() {
    this.type = 'subsegment';
    if (this.parent) {
        this.parent_id = this.parent.id;
    }
    if (this.segment) {
        this.trace_id = this.segment.trace_id;
    }
--  return JSON.stringify(this);
++ return this.serialize();
}; 
/**
 * Returns the formatted subsegment JSON string.
 */
Subsegment.prototype.toString = function toString() {
--  return JSON.stringify(this);
++ return this.serialize();
};

++ Subsegment.prototype.serialize = function serialize() {
++   return JSON.stringify(this);
++ };
```

This second option would have a smaller footprint and would still help us (Powertools) fix the issue, however I'd prefer the previous option so that all X-Ray customers can avoid seeing errors.

We would love to contribute the change directly back to the SDK so that it can benefit all customers and not just Powertools ones. If I were to open a PR, would be open to review it and help with the merge+release process once it's up to the project's standards?

dreamorosi:
*Issue #, if available:* closes #616

*Description of changes:*

This PR introduces a new method called `serialize` to the prototype of the `Segment` and `Subsegment` objects. This method wraps the `JSON.stringify()` call that was previously made by the `toString()` and `format()` methods of these two classes.

When calling the `JSON.stringify()` function, the method also passes a replacer function. This replacer for now includes an `if` statement that casts `BigInt` objects to `string`. This avoids runtimes errors described in the linked issue.

By extracting the method, customers can easily override it and extend the replacer to suit their needs.

In addition to the method, the PR includes also changes to the type definition and the addition of a test.

### DIFF
--- a/packages/core/lib/segments/attributes/subsegment.d.ts
+++ b/packages/core/lib/segments/attributes/subsegment.d.ts
@@ -61,6 +61,8 @@ declare class Subsegment {
   toString(): string;
 
   toJSON(): { [key: string]: any };
+
+  serialize(object?: Subsegment): string;
 }
 
 export = Subsegment;
--- a/packages/core/lib/segments/attributes/subsegment.d.ts
+++ b/packages/core/lib/segments/attributes/subsegment.d.ts
@@ -62,7 +62,7 @@ declare class Subsegment {
 
   toJSON(): { [key: string]: any };
 
-  serialize(object?: Subsegment): string;
+  serialize(subsegment?: Subsegment): string;
 }
 
 export = Subsegment;
--- a/packages/core/lib/segments/segment.d.ts
+++ b/packages/core/lib/segments/segment.d.ts
@@ -64,6 +64,8 @@ declare class Segment {
   format(): string;
 
   toString(): string;
+
+  serialize(object?: Segment): string;
 }
 
 export = Segment;
--- a/packages/core/lib/segments/segment.d.ts
+++ b/packages/core/lib/segments/segment.d.ts
@@ -65,7 +65,7 @@ declare class Segment {
 
   toString(): string;
 
-  serialize(object?: Segment): string;
+  serialize(segment?: Segment): string;
 }
 
 export = Segment;
--- a/packages/core/lib/segments/segment.js
+++ b/packages/core/lib/segments/segment.js
@@ -439,13 +439,10 @@ Segment.prototype.toString = function toString() {
 };
 
 Segment.prototype.serialize = function serialize(object) {
-  return JSON.stringify(object ?? this, (_, value) => {
-    if (typeof value === 'bigint') {
-      return value.toString();
-    }
-
-    return value;
-  });
+  return JSON.stringify(
+    object ?? this,
+    SegmentUtils.getJsonStringifyReplacer()
+  );
 };
 
 module.exports = Segment;
--- a/packages/core/lib/segments/segment.js
+++ b/packages/core/lib/segments/segment.js
@@ -431,11 +431,21 @@ Segment.prototype.format = function format() {
     false
   );
 
-  return JSON.stringify(thisCopy);
+  return this.serialize(thisCopy);
 };
 
 Segment.prototype.toString = function toString() {
-  return JSON.stringify(this);
+  return this.serialize();
+};
+
+Segment.prototype.serialize = function serialize(object) {
+  return JSON.stringify(object ?? this, (_, value) => {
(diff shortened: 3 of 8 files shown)


misterjoshua (commented):
Looks pretty good. I have a few suggestions.

packages/core/lib/segments/segment.js
```diff
@@ -431,11 +431,21 @@ Segment.prototype.format = function format() {
...
+};
+
+Segment.prototype.serialize = function serialize(object) {
+  return JSON.stringify(object ?? this, (_, value) => {
+    if (typeof value === 'bigint') {
+      return value.toString();
+    }
+
+    return value;
+  });
```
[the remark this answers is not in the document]
dreamorosi: I have extracted the replacer into the `SegmentUtils` object, which seems to hold shared utilities used by both `Segment` & `Subsegment`, and I also added an unit test to cover it.

Thanks for the suggestion!

misterjoshua (approved):
LGTM!

packages/core/lib/segments/attributes/subsegment.d.ts
```diff
@@ -61,6 +61,8 @@ declare class Subsegment {
   toString(): string;
 
   toJSON(): { [key: string]: any };
+
+  serialize(object?: Subsegment): string;
```
carolabadeer, line 65:
```suggestion
  serialize(subsegment?: Subsegment): string;
```

Better naming would be `subsegment` instead of `object`, which follows the naming convention used in the methods above as well

packages/core/lib/segments/segment.d.ts
```diff
@@ -64,6 +64,8 @@ declare class Segment {
   format(): string;
 
   toString(): string;
+
+  serialize(object?: Segment): string;
```
carolabadeer, line 68:
```suggestion
  serialize(segment?: Segment): string;
```
Same as comment above

carolabadeer (commented):
Thanks for making this change! Left a few small comments, but otherwise looks good

dreamorosi:
Hi @carolabadeer thanks for the review and the comments, I have committed both your suggestions!

carolabadeer (approved):

PR #641, opened 2024-04-18
TraceId timestamp supports string starts with '0'

mxiamxia:
*Issue #, if available:*
If upstream is sending X-Ray like trace id but with timestamp section starts with `0`, such as "1-0074e2eb-694001932729d04067bb5440"
**Actual,**
The bug is: the trace id is parsed to "1-74e2eb-694001932729d04067bb5440".
**Expected**
The expected behavior is: the trace id is parsed to "1-0074e2eb-694001932729d04067bb5440".

*Description of changes:*
If the timestamp section in X-Ray like trace id less than 8 digits, pad `0`s as prefix.

*Description of changes:*
Unit Test


vastin (approved):
LGTM. ty.

jj22ee (approved):

Issue #650, opened 2024-05-22
[fetch] does not support fetch options like proxy


Kruspe:
When using the `aws-xray-sdk-fetch` instrumentation it is not possible to supply further options to the fetch call like the agent to supply the undici ProxyAgent.

The `capturedFetch` call only calls the global available fetch call with one argument https://github.com/aws/aws-xray-sdk-node/blob/73e1fcaba509a6b6fcfc81254a25da37189bea4b/sdk_contrib/fetch/lib/fetch_p.js#L117 and does not pass down options provided by the user.

I would propose to add a third option to the function that patches the global fetch. This would allow the user to use all fetch options and additionally supply their own segment information if required.
https://github.com/aws/aws-xray-sdk-node/blob/73e1fcaba509a6b6fcfc81254a25da37189bea4b/sdk_contrib/fetch/lib/fetch_p.js#L59

For this change it would be necessary to think about how not to introduce a breaking change for users that expect the second argument to be used for segment information. But I guess we can write some internal logic to check the supplied args.

Once I receive some feedback I can get started on a PR and implement the changes.

jj22ee:
Hi, you should be able to pass options into the fetch call. For example, the following code sample will change the method type used in fetch from GET to POST. This will be reflected in the trace as well.

```
const { captureFetchGlobal } = require('aws-xray-sdk-fetch');
const fetch = captureFetchGlobal();

...

  const result = fetch('https://someURL.com', {
    method: "POST",
    headers: {
      "Content-Type": "application/json",
    },
    body: JSON.stringify({data:123, data2:"234"}),
  });
```

The options are added into the fetch call through the `Request` object: https://github.com/aws/aws-xray-sdk-node/blob/73e1fcaba509a6b6fcfc81254a25da37189bea4b/sdk_contrib/fetch/lib/fetch_p.js#L63-L117

Kruspe:
Hey, what you say is true but `Request` does not support the dispatcher option as far as I know.

With fetch you can set the dispatcher and the request will go through the Proxy.
```js
import { ProxyAgent } from 'undici';

fetch("http://localhost", {
  dispatcher: new ProxyAgent("http://localhost:4444"),
}));
```

If we create a request with similar options like this:
```js
import { ProxyAgent } from 'undici';

const request = new Request("http://localhost", {
  dispatcher: new ProxyAgent("http://localhost:4444"),
});

fetch(request)
```
[...]

PR #653, opened 2024-06-05
Pass dispatcher options to global fetch

Kruspe:
*Issue: #650

*Description of changes:*
As the global fetch in node follows the [undici fetch](https://github.com/nodejs/undici?tab=readme-ov-file#undicifetchinput-init-promise) implementation there is an extra dispatcher option that can be passed but is not supported by the Request class.
We check if the extra option is supplied and if so pass the option directly to the global fetch call. All other paths already pass all options as expected.

### DIFF
--- a/sdk_contrib/fetch/lib/fetch_p.js
+++ b/sdk_contrib/fetch/lib/fetch_p.js
@@ -63,6 +63,10 @@ const enableCapture = function enableCapture(baseFetchFunction, requestClass, do
     const request = typeof args[0] === 'object' ?
       args[0] :
       new requestClass(...args);
+    let fetchOptions = undefined;
+    if (args[1] && 'dispatcher' in args[1]) {
+      fetchOptions = { dispatcher: args[1].dispatcher };
+    }
 
     // Facilitate the addition of Segment information via the request arguments
     const params = args.length > 1 ? args[1] : {};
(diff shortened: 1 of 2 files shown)


sdk_contrib/fetch/lib/fetch_p.js
```diff
@@ -116,7 +120,7 @@ const enableCapture = function enableCapture(baseFetchFunction, requestClass, do
       const requestClone = request.clone();
       let response;
       try {
-        response = await baseFetchFunction(requestClone);
+        response = await baseFetchFunction(requestClone, fetchOptions);
```
jj22ee, line 123: Could this be changed to `baseFetchFunction(...args);`? This way the inputs to the fetch function are as what was originally intended as if there were no instrumentation at all. There would be no need to create `fetchOptions` either.

jj22ee:
The tests are failing on NodeJS v20.x. Can you take a look into them, perhaps my suggestion could fix them?

Kruspe:
Hey sorry for the late update! I did not get a notification that the tests ran here.
An update to Node 20 seems to have broken the previous implementation but your suggestion should work just fine. I'll keep a close look now but my guess would be that all the tests should now pass inside the GH actions.

jj22ee (approved):

Issue #655, opened 2024-06-06
Cannot set properties of undefined (setting 'namespace') 3.7.0 error


daanootje:
Could be a duplicate of #654. After recently changing to 3.7.0 we are encountering the following error:

```
2024-06-06T10:52:44.066Z | INIT_START Runtime Version: nodejs:18.v29 Runtime Version
2024-06-06T10:52:44.895Z | undefined ERROR _X_AMZN_TRACE_ID is missing required information
2024-06-06T10:52:44.896Z | undefined ERROR Empty or non-string trace ID provided
2024-06-06T10:52:45.750Z | START RequestId: 60bee989-dc24-514d-a1cc-b7524b09db45
2024-06-06T10:52:46.105Z	| ERROR	Invoke Error 	
{
    "errorType": "RequestError",
    "errorMessage": "Cannot set properties of undefined (setting 'namespace')",
    "code": "ERR_GOT_REQUEST_ERROR",
    "name": "RequestError",
    "stack": [
        "RequestError: Cannot set properties of undefined (setting 'namespace')",
        "    at captureOutgoingHTTPs (/var/task/node_modules/aws-xray-sdk-core/dist/lib/patchers/http_p.js:108:30)",
        "    at Object.captureHTTPsRequest [as request] (/var/task/node_modules/aws-xray-sdk-core/dist/lib/patchers/http_p.js:175:16)",
        "    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)"
    ]
}
2024-06-06T10:52:46.109Z | END RequestId: 60bee989-dc24-514d-a1cc-b7524b09db45 
2024-06-06T10:52:46.109Z | REPORT RequestId: 60bee989-dc24-514d-a1cc-b7524b09db45	Duration: 359.46 ms	Billed Duration: 360 ms	Memory Size: 1024 MB	Max Memory Used: 137 MB	Init Duration: 1683.30 ms	
XRAY TraceId: 1-666194fb-4d3d1850fcf803f9e905fdf7	SegmentId: 161caf1c19d1cbf2	Sampled: true	
```

The code tries to bind the xray trace id to an outgoing http request.

```typescript
import * as awsXRay from 'aws-xray-sdk-core'
import https from 'https'

const enableTracing = true

export function makeXrayHttps({ enableTracing }: { enableTracing: boolean }) {
    return enableTracing ? awsXRay.captureHTTPs(https, true) : https // <-- Most likely failing here
}

[...]

Issue #654, opened 2024-06-06
3.7.0 fails reading 'addAttribute' of undefined


tenjaa:
Since the update to 3.7.0 my lambda fails.

Here are the logs of it failing:
```
INIT_START Runtime Version: nodejs:18.v29	Runtime Version ARN: arn:aws:lambda:eu-west-1::runtime:e78f4a8a086bd7395a083454a3b1a85e8bb06f09428b715c8078dfc31ce0c35c
2024-06-06T10:01:18.236Z	undefined	ERROR	_X_AMZN_TRACE_ID is missing required information
2024-06-06T10:01:18.237Z	undefined	ERROR	Empty or non-string trace ID provided
2024-06-06T10:01:18.341Z	undefined	INFO	Found <XXX>. Setting up proxy.
START RequestId: <XXX> Version: 453
2024-06-06T10:01:18.469Z	<XXX>	ERROR	Invoke Error 	
{
    "errorType": "TypeError",
    "errorMessage": "Cannot read properties of undefined (reading 'addAttribute')",
    "stack": [
        "TypeError: Cannot read properties of undefined (reading 'addAttribute')",
        "    at file:///var/task/index.mjs:7438:18",
        "    at file:///var/task/index.mjs:16411:16",
        "    at file:///var/task/index.mjs:11253:14",
        "    at async file:///var/task/index.mjs:8931:26",
        "    at async Catalog.getCredentials (file:///var/task/index.mjs:36684:26)",
        "    at async Runtime.handler (file:///var/task/index.mjs:36631:7)"
    ]
}

END RequestId: <XXX>
REPORT RequestId: <XXX>	Duration: 114.86 ms	Billed Duration: 115 ms	Memory Size: 512 MB	Max Memory Used: 85 MB	Init Duration: 399.98 ms	
XRAY TraceId: 1-<XXX>	SegmentId: 0b2b67f75caf1811	Sampled: true	
```

This is how it looked right before the update:
```
INIT_START Runtime Version: nodejs:18.v29	Runtime Version ARN: arn:aws:lambda:eu-west-1::runtime:e78f4a8a086bd7395a083454a3b1a85e8bb06f09428b715c8078dfc31ce0c35c
2024-06-06T10:01:42.058Z	undefined	INFO	Found proxy.resources.vwapps.cloud. Setting up proxy.
START RequestId: <XXX> Version: 451
END RequestId: <XXX>
[...]

PR #659, opened 2024-06-06
Add workflow to deprecate SDK version

jj22ee:
*Issue #, if available:*

*Description of changes:*
Add workflow to specifically Deprecate v3.7.0 of the SDK. In the future, this workflow can be made more generic (add deprecation version and message as workflow input), but for now, 3.7.0 is specified. 

Similar deprecation workflow: https://github.com/aws-observability/aws-rum-web/blob/main/.github/workflows/npm_deprecate.yaml

### DIFF
--- a/.github/workflows/deprecate_version.yml
+++ b/.github/workflows/deprecate_version.yml
@@ -1,0 +1,31 @@
+name: Deprecate X-Ray Node SDK Version
+on:
+  workflow_dispatch:
+
+jobs:
+  deprecate_xray_node_sdk_version:
+    name: Deprecate X-Ray Node SDK version in NPM registry
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+
+      - run: npm install -g npm@8.19.4
+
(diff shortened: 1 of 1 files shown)


wangzlei (approved):

.github/workflows/deprecate_version.yml
```diff
@@ -0,0 +1,31 @@
...
+        uses: actions/setup-node@v1
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+
+      - run: npm install -g npm@8.19.4
+
+      - name: Deprecate Version 3.7.0
+        run: |
+          npm deprecate aws-xray-sdk@3.7.0 "3.7.0 is deprecated due to known issue in Lambda"
```
srprash, line 20: Is there an issue that we can link to in the deprecation message for further clarity?

srprash: I see a couple here: 
https://github.com/aws/aws-xray-sdk-node/issues/654
https://github.com/aws/aws-xray-sdk-node/issues/655

srprash: Are these not why we are deprecating 3.7.0?

jj22ee: We found the issue locally first, but you are right, these are the same issue.

PR #651, opened 2024-05-31
[Lambda] Replace Facade with No-Op if trace header is missing data

majanjua-amzn:
To support Root Lambda PassThrough, we want to have a no-op segment when PassThrough mode is enabled to allow for SDK to have its functionalities disabled for the given segment. The following tests were done using two lambda functions, A and B, where A calls B.

Lambda A PassThrough calling Prod Lambda B Active:
- A: `Root=1-665a0c0c-2420a39e4fca953752e2e9f4;Lineage=061379e8:0`
- B: `Root=1-665a0c0c-2420a39e4fca953752e2e9f4;Parent=8c885bca7ccc2033;Sampled=1;Lineage=061379e8:0|cbc84ad3:0`

Lambda A Active calling Prod Lambda B Active:
- A: `Root=1-665a39a0-7e53f00f3b1da8ad70666729;Parent=ce7fe8bbc2783986;Sampled=1;Lineage=061379e8:0`
- B: `Root=1-665a39a0-7e53f00f3b1da8ad70666729;Parent=e80ec591cab089eb;Sampled=1;Lineage=061379e8:0|cbc84ad3:0`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #660, opened 2024-06-07
Lambda PassThrough trace header support

majanjua-amzn:
Lambda trace header with passthrough will be missing parent and sampled attributes, we should be able to support this.

#651 was reverted and this PR works to do the same in a simpler way.

*Description of changes:*
- Detect lambda trace context and mark segments with `noOp` attribute, signifying passthrough mode
- Propagate this attribute to subsegments when they are added
- Detect `noOp` attribute to stop trace context propagation in passthrough mode
- Unit test checking if segment generated with passthrough behaviour (`segment.noOp == true`) results in the trace header not being propagated as expected

Testing:
Tested appropriate behaviour using two lambda functions:
- Lambda A (passthrough) calls Lambda B (active): [x]
- Lambda A (active) calls Lambda B (active):

And tested combinations of the following:
- Lambda function with old passthrough behaviour
- Lambda function with new passthrough behaviour
- Lambda function with AWS SDK v2
- Lambda function with AWS SDK v3

### DIFF
--- a/packages/core/lib/utils.js
+++ b/packages/core/lib/utils.js
@@ -168,6 +168,10 @@ var utils = {
       if (!traceData) {
         traceData = {};
         logger.getLogger().error('_X_AMZN_TRACE_ID is empty or has an invalid format');
+      } else if (traceData.root && !traceData.parent && !traceData.sampled) {
+        // Lambda PassThrough only has root, treat as valid in this case and mark the segment
+        segment.noOp = true;
+        valid = true;
       } else if (!traceData.root || !traceData.parent || !traceData.sampled) {
         logger.getLogger().error('_X_AMZN_TRACE_ID is missing required information');
       } else {
--- a/packages/core/lib/patchers/aws3_p.ts
+++ b/packages/core/lib/patchers/aws3_p.ts
@@ -140,7 +140,9 @@ const getXRayMiddleware = (config: RegionResolvedConfig, manualSegment?: Segment
     }
   }
 
-  args.request.headers['X-Amzn-Trace-Id'] = traceHeader;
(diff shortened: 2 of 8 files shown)


wangzlei (approved):
LGTM

jj22ee (commented):
There is one more patcher (`fetch`) that is introduced in sdk_contrib folder, can we confirm if your changes need to be applied to this one as well.
https://github.com/aws/aws-xray-sdk-node/blob/master/sdk_contrib/fetch/lib/fetch_p.js#L107-L110

packages/core/lib/utils.js
```diff
@@ -168,6 +168,10 @@ var utils = {
       if (!traceData) {
         traceData = {};
         logger.getLogger().error('_X_AMZN_TRACE_ID is empty or has an invalid format');
+      } else if (traceData.root && !traceData.parent && !traceData.sampled) {
+        // Lambda PassThrough only has root, treat as valid in this case and mark the segment
+        segment.noOp = true;
+        valid = true;
```
jj22ee, line 174: If in `"valid no-op"` scenario, do we want to ensure that the segment/subsegment is not being flushed to XRay Daemon/FluxPump, and if so, is this ensured in this PR?

jj22ee: https://github.com/aws/aws-xray-sdk-node/blob/dd12415098a5cb4d9846683cae9508e857fa8835/packages/core/lib/utils.js#L171-L190
- If `no-op`, then `notTraced` is true, which means that segment is not sent

jj22ee (approved):
Thanks!

Issue #679, opened 2024-09-17
fetch calls with anything but a request object will execute without the x-amzn-trace-id header


berndfuhrmann:
Consider this code
```
import AWSXRay from 'aws-xray-sdk-core';
import { captureFetchGlobal } from 'aws-xray-sdk-fetch';

captureFetchGlobal();

var segment = new AWSXRay.Segment('test');
var ns = AWSXRay.getNamespace();

ns.run(async function () {
  AWSXRay.setSegment(segment);
  fetch('http://localhost:3000')
  segment.close();
});
```
I would expect `localhost:3000` to receive the `x-amzn-trace-id` header, but it doesn't.
When changing the fetch call like this: `fetch(new Request("http://localhost:3000"));` it works.
Working on a PR to fix this.

berndfuhrmann:
Thanks a lot for fixing the codecov issue and merging this.
Keen to see this released :-)

PR #687, opened 2024-09-20
Fix missing headers for fetch

berndfuhrmann:
*Issue #679:*

*Description of changes:*
* For integration tests, added a small HTTP server that receives requests made via fetch. This offers the ability to check headers that were actually supplied in the request.
* I reverted a changed introduced recently, so X-Amzn-Trace-Id headers are sent again
* Exporting `enableCapture` function. This provides the ability to use this without changing globals. It also provides the ability to use different Request classes, e.g. undici.Request instead of globalThis.Request. This should provide a way for anyone who wants to use undici with `dispatcher`.


berndfuhrmann:
Also added this bug to undici:
https://github.com/nodejs/undici/issues/3630

jj22ee:
Thanks for the fix and improvements for this and the undici package!

The changes lgtm. We can remove `package-lock.json` in fetch package as this is a monorepo, I suppose your changes to `package.json` should affect the root `package-lock.json`  instead.

Not sure why the codecov is failing suddenly on the Node16.x/Ubuntu test, I'll look into this.

jj22ee (approved):
LGTM 🚢

jj22ee:
I've fixed the failing codecov in https://github.com/aws/aws-xray-sdk-node/pull/692 and merged the fix into your branch, along with fixing the package-lock files.

lennart-land:
@jj22ee with this PR our production code broke, since we use this behind a proxy.
May i ask why the test `resolves to response through proxy when fetch options are supplied` and also the implementation to support proxies was removed?
At least a minor version release does not look correct here, since it introduced a breaking change to us.

berndfuhrmann:
@lennart-land
The problem with test was this line:
`stubFetch.should.have.been.calledOnceWith(request, {dispatcher: proxyStub});`
Expecting this makes the test brittle, because the way adding the header works is to clone the request. When doing that, the dispatcher is "baked" into that request object, so stubFetch would just be called with that request. It works after a fix in undici.
This was the relevant fix on their side: https://github.com/nodejs/undici/issues/3630#event-14342711199
So with that undici patch, things should work. However, that patch seems to be included only in undici 7, which is in alpha stage currently. That is unfortunate timing.

jj22ee:
Hi @lennart-land , the previous change to introduce support for proxies in fetch introduced a regression (X-Amzn-Trace-Id header not propagated to downstream when not using a request object in Fetch), so we reverted the change in this PR. That said, you are right that a minor version release was incorrect, that is an error on my end.

For now, the current workaround is to pin `aws-xray-sdk-fetch` to `v3.10.1`. In the mean time, when the undici patch is released, that is an alternative option. We'll also look into an alternative solution for proxy support that doesn't break X-Amzn-Trace-Id propagation.

jj22ee:
Re-opening https://github.com/aws/aws-xray-sdk-node/issues/650

PR #696, opened 2024-11-17
Add proxy support for global fetch

Kruspe:
*Issue #650:*

*Description of changes:*
Support setting the dispatcher for the global fetch implementation.
Instead of ignoring the created request, we now use it and pass the dispatcher option if it is available.

Added an integration test with a reverse proxy to check that the trace header is there.

### DIFF
--- a/sdk_contrib/fetch/lib/fetch_p.js
+++ b/sdk_contrib/fetch/lib/fetch_p.js
@@ -77,6 +77,7 @@ function enableCapture(baseFetchFunction, requestClass, downstreamXRayEnabled, s
 
     // Facilitate the addition of Segment information via the request arguments
     const params = args.length > 1 ? args[1] : {};
+    const fetchOptions = 'dispatcher' in params ? {dispatcher: params.dispatcher} : undefined;
 
     // Short circuit if the HTTP is already being captured
     if (request.headers.has('X-Amzn-Trace-Id')) {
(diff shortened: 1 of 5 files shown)


sdk_contrib/fetch/lib/fetch_p.js
```diff
@@ -127,7 +128,7 @@ function enableCapture(baseFetchFunction, requestClass, downstreamXRayEnabled, s
       const requestClone = request.clone();
       let response;
       try {
-        response = await baseFetchFunction(requestClone);
+        response = await baseFetchFunction(requestClone, fetchOptions);
```
Kruspe, line 131: My first idea was to pass down the entire args again but this would overwrite the changes we do to the request like adding the trace header as the headers from the options take priority over the headers that are already set for a request.

berndfuhrmann: Consider this code:
```
const agent = new Agent({maxSockets: 1234});
const request = new Request('someurl', { dispatcher: agent });
fetch(request);
```
Here, fetch is only called with one parameter. The dispatcher is "baked in" into the request. For undici, it's actually a symbol property. When you call `baseFetchFunction(requestClone, fetchOptions)`, `requestClone` doesn't contain that property, So that per-request-setting wouldn't be passed on.

berndfuhrmann: Made an attempt to fix this. Sorry for the complexity, but it should work more reliably for you.
https://github.com/aws/aws-xray-sdk-node/pull/698

Kruspe: #698 fixes a different issue that is there so I would open another issue to point to your PR to.

Issue #648, opened 2024-05-14
fetch with URL object fails with `Cannot read properties of undefined (reading 'has')`


borgoat:
We are using `fetch` passing a URL object as `resource` parameter - and a second `options` argument containing headers. The first parameter may be any object with a stringifier.[^1]

However, currently it implies here[^2] that any `object` passed as first argument to `fetch` must be a `Request`.

```
TypeError: Cannot read properties of undefined (reading 'has')\n    at o9.fetch (/node_modules/.pnpm/aws-xray-sdk-fetch@3.6.0_aws-xray-sdk-core@3.6.0/node_modules/aws-xray-sdk-fetch/lib/fetch_p.js:71:25)\n    at o9.getData (/libs/instruments/service-cityfalcon/src/lib/cityfalcon.service.ts:179:12)\n    at params (/libs/instruments/service-cityfalcon/src/lib/cityfalcon.service.ts:60:14)\n    at Array.map (<anonymous>)\n    at o9.getMarketNewsTimeline (/libs/instruments/service-cityfalcon/src/lib/cityfalcon.service.ts:59:18)\n    at mWe (/libs/instruments/cloud/src/lib/market-news/market-news.get-market-news.fn/index.ts:103:12)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at gfe (/node_modules/.pnpm/@middy+core@4.7.0/node_modules/@middy/core/index.js:120:32)
```

[^1]: https://developer.mozilla.org/en-US/docs/Web/API/fetch
[^2]: https://github.com/aws/aws-xray-sdk-node/blob/58dd586b2e1de562dce07c74e0d066d2b74ad210/sdk_contrib/fetch/lib/fetch_p.js#L63C11-L65

rslfilho:
Im having the same issue here, was using a URL object as first parameter for `fetch`, and encounter the same error. Not every time the first parameter is a object, it will be a `Request`. For now i updated my code to send the `URL.href` string as `resource`, but should be fixed

tbartley:
This also impacts AWS Amplify as well because it invokes fetch with URLs

PR #697, opened 2024-11-18
fix 648 url object

berndfuhrmann:
*Issue #, if available:*
#648
*Description of changes:*
Implementation:
`args[0]` in fetch_p.js, line 74 could be any stringifyable object, not just a string. That object should be converted to a string and used as URL. The only exception to this is if `args[0]` is already a request object.
This PR assumes that such a request object must be from the same implementation as the one given by requestClass. So if you would put a one Request object from e.g. node-fetch into the native fetch implementation, this will stringify that request object.
I couldn't find any "official" way to distinguish between a request object and a stringifyable object. But you're supposed to call fetch with the corresponding request implementation, I think this is fair game.

For that reason, the test was changed, so it will provide the native Request object.


gabrielmartinez2004:
Do we have any updates here?

berndfuhrmann:
This has been untouched for month. @jj22ee, not sure, is there anything I can do?

PR #698, opened 2024-11-18
Fix 648 url object not accepted and 650 dispatcher not respected

berndfuhrmann:
*Issue #, if available:*
#648, #650
*Description of changes:*
There is a seperate PR for #648, see description there.
For #650 when fetch is instrumented, we check now if there is a dispatcher property settable on a request which is not transferred to the cloned request. It is a bug detector for undici. If that bug is detected, a work-around is enabled, which will search for the dispatcher symbol property and copy it manually to the new request. A test for that has been added.
In the future, this work-around can be removed, once the fix in undici is sufficiently available. For now, this should resolve @Kruspe problem.
Please note that the bug detection is only done when `enableCapture` is called, so only when the program initializes. This minimizes impacts to performance.
I included my fix to #648, since not doing that would make a difference in how the unit test works.

### DIFF
--- a/sdk_contrib/fetch/lib/fetch_p.js
+++ b/sdk_contrib/fetch/lib/fetch_p.js
@@ -71,7 +71,7 @@ function enableCapture(baseFetchFunction, requestClass, downstreamXRayEnabled, s
     const thisDownstreamXRayEnabled = !!downstreamXRayEnabled;
     const thisSubsegmentCallback = subsegmentCallback;
     // Standardize request information
-    const request = typeof args[0] === 'object' ?
+    const request = args[0] instanceof requestClass ?
       args[0] :
       new requestClass(...args);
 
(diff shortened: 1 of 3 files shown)


Kruspe:
I'd say this is unrelated to #650 and is rather a new issue to fix the support for Requests that have a dispatcher set directly.

#650 aims at fixing the behavior for the following calls to fetch:
```js
fetch("https://something.com", {dispatcher: agent})
```

berndfuhrmann:
@Kruspe 
This is from a test I added: `await activeFetch('https://www.foo.com', { dispatcher: agent});`. Isn't that what you're talking about?

Kruspe:
Ah I see what you mean. True the newly created Request would contain this and copying this over correctly would fix that issue.

Scenarios like this:
```js
fetch(new Request("https://something.com"), {dispatcher: agent});
```
would still be broken though right?

In general I think it's a good idea to combine the options into the calls we pass to the original fetch as otherwise we would be losing those options.

berndfuhrmann:
@Kruspe Thanks for pointing that out. Added a test and a fix.

thomaswr:
What's the current status on that?

PR #748, opened 2026-03-19
fix(daily-scan): move DependencyCheck install to /tmp to avoid self-scan

thpierce:
DependencyCheck is unzipped into the repo checkout directory, then scans `.` which includes its own `dependency-check/lib/*.jar` files. This causes false positive CVE findings (commons-beanutils, gson, h2, httpclient5, logback-core, etc.) on every run.

This PR moves the unzip to `/tmp` so the tool binaries are outside the scan path.

Ref: V1902405376


thpierce:
Incorrect fix

